PR #15, opened 2015-02-20
Update of SPECS cmor version before general cleaning of the branches for CMOR3

pabretonniere:
Cleaning in the code, adding a few late modifications and updating it to the latest version from the official master branch (2.9.1)


dnadeau4:
@pabretonniere Does  this still make any sense for cmor 3.2?  There is a lot of code changes and it is impossible to merge with the latest cmor version.

PR #30, opened 2015-06-30
Add support for --with-hdf5 option to configure

tfeb:
The `--with-hdf5` option was previously documented in `INSTALL` but not actually supported: this pull request makes it be supported.  It looks enormous but that's because there's a new generated copy of `configure`, which is enormous: the actual changes are not as big as all that!

I don't understand the build process any better than I needed to so this probably should be reviewed fairly carefully.  I have built CMOR with this configuration however, and with HDF5 in an non-standard place.

I am fairly sure that building CMOR doesn't actually need the HDF5 libraries, but I think it may well depend on the headers, so it needs to get the right ones.  Building programs which link with CMOR may well need the libraries, and will also need to get the right ones.

I changed the following files:
- `configure.ac`: add support (see note 1 below)
- (`configure` is generated from `configure.ac` and committed, see note 3 below)
- `Makefile.in`: add support (see note 2 below)
- `compile_line.txt.in`: add support
- `setup.py.in`: add support

## Note 1
`configure.ac` has a bunch of assignments setting `LIB=""` which I think are spurious and which I've removed: if they're *not* spurious they should be put back!

## Note 2
`Makefile.in` is not in a good state: this change hasn't made things *worse* but I haven't tried to fix any of it.

## Note 3
The version of autotools I have seems to be older than the one that CMOR's configure was previously built with: it might be worth recreating it with a newer one.


doutriaux1:
hum.. I will have to double check, but I don't think hdf5 flags are needed any longer since I'm getting the flags via 
```
nc-config --cflags
```
Let me double check before approving.

Did you have problems installing w/o them?

tfeb:
Doh.  I didn't know about `nc-config`, but it looks like that is the answer and my change is not needed.

(The reason I hadn't understood this was because I'm trying to sort out how to get CMOR installed on a 'naked' machine, while working on a system which already has an annoying (for these purposes) amount of stuff installed in central locations, making it hard to know when things are depending on (for instance) a centrally-installed version of NetCDF.

So I don't think this is needed, sorry.

If you like I'll submit a much smaller request to remove the text in `INSTALL` saying it is needed!

doutriaux1:
If you don't mind, that would be great. Also I did push some changes to the makefile.in last week, so when running configure you should get a cleaner Makefile now.

tfeb:
Cool, I'll do that tomorrow (UK time) and close this one in the meantime.

PR #45, opened 2015-10-12
Use OS UUID library

tfeb:
Hi, I made some changes to CMOR to allow it to use the OS-provided uuid library rather than the OSSP one.  This means there is one less dependency on Linux, which is especially good since the OSSP stuff seems to be semi-unavailable (the ftp server has been down for months: there are still mirrors, for now).

I am pretty sure this works (with the caveats below), but I'm submitting a pull request mostly so I can stop worrying about keeping my branch up to date: if you don't want to take them that's fine, I'm happy to go back to the OSSP stuff.

A couple of notes:

* I have not tested this on anything but RHEL: I think it should work on any reasonable Linux however.  But if there is a requirement to run on platforms which are not Linux it's probably worth checking that they either have, or can have, a compatible uuid library.
* It is not extensively tested even on Linux: caveat emptor!  All the changes are to `Src/cmor.c`: someone should probably check the diffs.
https://guides.github.com/features/mastering-markdown/


doutriaux1:
thx @tfeb  @dnadeau4 and I will test on various platforms and then will merge soon.

doutriaux1:
quick comment we might need some pre-processor code in there to switch back to old code if uuid is not on system.

tfeb:
@doutriaux1: I thought about that overnight.  I think an equivalent but better approach is to allow it to compile against a version of the uuid library which isn't installed in the standard location.  So then you could fetch, build and install the GNU uuid, and just use an option to configure to get it to work.

I *think* that ought to work now, in fact, but I have not tested.

If that sounds like a good answer I will check it does actually work, and add a note to `INSTALL` as to where to obtain a compatible library.  I'll try to do that today (UK time).

doutriaux1:
@tfeb isn't it sort of what is in already? That what the --with-uuid flags to configure are for as far as I remember.

tfeb:
Yes it pretty much is just that: I just wanted to check it actually worked and find a canonical location for the library sources to note in the INSTALL file before dumping it in your lap as it were.

(Sorry, it will also be tomorrow now, I've been disorganised.)

tfeb:
OK, I have noted where the uuid library comes from (util-linux) and that it seems to be compatible across at least several platforms.

What I haven't managed to do is to build a version of util-linux completely from source (rather: I have, but installing it needs root permissions because there are some SUID bits, unrelated to uuids of course).  So that might potentially be a problem, although almost certainly it will be present (or easily installable) on any Linux distro.

dnadeau4:
I am still using OSSP uuid, @doutriaux1 do you agree to change to this more standard version.

dnadeau4:
@tfeb I changed the code to use the "uuid" from the systems (MAC/Linux), but I have conflicts with UV-CDAT and cdms2.  It is easier for me just to add ossuuid library in the python/lib directory than changing cdms2 on Linux and the MAC and it will take to much time to coordinate everything right now.  I have the branch ready, and will do this with @doutriaux1 in a further release.    Thanks for your help.

tfeb:
@dnadeau4 that's fuine of course.  Do you want me to close the pull request?

tfeb:
Closing this to remove it from my dashboard...

PR #46, opened 2015-10-21
tracking_prefix now filtered so it only applies if CMIP6 project_id is set

TobiasWeigel:
As agreed in the final version of the PID services white paper, the tracking_prefix will now only be written if the project_id is set to (exactly) "CMIP6" to prevent accidental use of the prefix in other projects.

Successfully tested by Jörg against CORDEX data with modified cmor tables.

### DIFF
--- a/Src/cmor.c
+++ b/Src/cmor.c
@@ -35,6 +35,8 @@ int cleanup_varid=-1;
 
 const char CMOR_VALID_CALENDARS[CMOR_N_VALID_CALS][CMOR_MAX_STRING] = { "gregorian","standard", "proleptic_gregorian","noleap","365_day","360_day","julian","none"};
 
+const char cmor_tracking_prefix_project_filter[CMOR_MAX_TRACKING_PREFIX_PROJECT_FILTER][CMOR_MAX_STRING] = { "CMIP6" };
+
 cmor_dataset_def cmor_current_dataset;
 cmor_table_t cmor_tables[CMOR_MAX_TABLES];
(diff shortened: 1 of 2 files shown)


TobiasWeigel:
Updated. Now checks against const string array defined in cmor.h. Not tested yet; will do so beginning of January.

TobiasWeigel:
Some more changes after testing this since the prefix was not ultimately written to output files. Now it is written; I've also thoroughly tested the combinations of project_id and tracking_prefix to find out whether it works as expected and it does (e.g. will complain when project_id matches, but no prefix is set; will ignore tracking_prefix if project_id is not in filter etc.). Please merge :-)

doutriaux1:
@dnadeau4 that looks good to me. If you are fine with it let's merge in.

dnadeau4:
We need to talk to Karl about this.  He was mentioning about not having 
DOI as a tracking_id prefix for CMIP6

It looks good to me, but we need to make sure he approves.

Denis

dnadeau4:
Tobias,  I have approved and merged your pull request. Could you make another request for the CMOR3 branch?  This will be the one used for CMIP6.

TobiasWeigel:
Hi guys :-) Yes, there was another discussion last week that Karl initiated.. it was finally resolved by agreeing that we have "21.14100/<UUID>" as the ID syntax. But it's never wrong to check again and confirm what we are doing here is correct.
I'll open another PR for the other branch immediately. Also note that at some point we'll have to put the actual tracking_prefix entry into the tables.

PR #48, opened 2016-01-23
Initial work on implementing cmake build system for cmor

chaosphere2112:
This is the initial stub of a CMAKE infrastructure for CMOR; it doesn't even remotely work yet, but should be a good guidepost for getting something functional. Right now, I was focused on trying to get wget to install correctly; it's very nearly there, running into a configure issue (which is probably something not getting set in the deps.cmake file that should be). After I get my HPSS to ESGF thing out I can dedicate a little more effort to this.


doutriaux1:
@chaosphere2112 I spoke about this with @taylor13 we're not sure that's the right thing to do. Please hold on on this.

aashish24:
@doutriaux1 happy to provide feedback or suggestions if you need help.

doutriaux1:
sure. but it's more a matter of priorities here.

taylor13:
I'm glad Sam is interested.  I think Sam, Charles, and maybe Denis, and I should should meet briefly early next week to discuss what the priorities are.  We didn't get complaints (as far as I recall) from the CMIP modelling groups indicating that CMOR was difficult to build.  It seems that the anecdotal complaints about CMOR were about building netCDF.  Literally thousands of users have downloaded netCDF files from PCMDI and I haven't heard any complain about building netCDF, so we need to evaluate what's the best use of resources here.  In the end, we might want to include as *one* option something along the lines of what Sam has outlined.  Again it's great that some thought is going into this.  Talk to you all Monday.

durack1:
@chaosphere2112 @doutriaux1 @dnadeau4 @taylor13 I do think it would be worthwhile to build in some automagical `travis` testing, just to make sure that merged commits aren't causing problems - during the testing phase it may also be useful to use the bot network to also test this on an array of different `*nix` and `OS X` systems..

doutriaux1:
@durack1 I agree that it would be a better use of resources.

doutriaux1:
@chaosphere2112 you added way too many dependencies here anyway. mpi, libcdms, etc... They have nothing to do with CMOR. If you want these you need to have it as pre-built. Or you should use UV-CDAT to install it.

chaosphere2112:
@doutriaux1 I just followed the dependency tree of the CMOR package in UV-CDAT; didn't spend too much effort working on this, just tried to get something stubbed out.

chaosphere2112:
@doutriaux1 @taylor13 @durack1 I'm booked all this week for another project, just wanted to lend some kind of a hand to @dnadeau4 since he's sweating about the deadline.

dnadeau4:
Since we now have anaconda installation for cmor, we don't really need cmake anymore.

aashish24:
> Since we now have anaconda installation for cmor, we don't really need cmake anymore.

@dnadeau4 I would suggest we move away from configure scripts that cmor has. You may want to use cmake for the building the cmor package itself (and not its dependencies) and conda for package management (which then can call cmake internally). Just my 2 cents :smile:

aashish24:
just to clarify -- 

use conda instead of cmake-superbuild 
use cmake instead of configure

dnadeau4:
There won't be any cmake.  

I put the installation procedure in the documenation [cmor.llnl.gov](documenation).

aashish24:
I guess my point was not understood: In conda recipe, we have the text as shown below which means that we are using configure scripts which is platform specific and hard to maintain). 

export CFLAGS="-Wall -g -m64 -pipe -O2  -fPIC"
export CXXLAGS="${CFLAGS}"
export CPPFLAGS="-I${PREFIX}/include"
export LDFLAGS="-L${PREFIX}/lib"
CONDA_LST=`conda list`
if [[ ${CONDA_LST}'y' == *'openmpi'* ]]; then
    export CC=mpicc
    export CXX=mpicxx
    export LC_RPATH="${PREFIX}/lib"
    export DYLD_FALLBACK_LIBRARY_PATH=${PREFIX}/lib
fi


./configure \
    --with-python=${PREFIX}   \
    --with-uuid=${PREFIX} \
    --with-udunits2=${PREFIX} \
    --with-netcdf=${PREFIX} \
    --with-libjson-c=${PREFIX} \
    --prefix=${PREFIX}
make 
make install

Issue #67, opened 2016-06-16
create_output_path doesn't work from Python


ehogan:
`create_output_path` is available in `pywrapper.py`, but `cmor._cmor` has no attribute `create_output_path`:
```
% ipython
Python 2.7.6 (default, Jan  3 2014, 16:42:21) 
Type "copyright", "credits" or "license" for more information.

IPython 2.2.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import cmor

In [2]: cmor.create_output_path('')
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-3-fa2c46fcc4fc> in <module>()
----> 1 cmor.create_output_path('')

/home/h01/ehogan/software/cmor/lib/python2.7/site-packages/cmor/pywrapper.pyc in create_output_path(varid)
    912     Returns: the path where the file will be created
    913     """
--> 914     return _cmor.create_output_path(varid)

AttributeError: 'module' object has no attribute 'create_output_path'

In [3]: 
```

dnadeau4:
It seems that this has been removed from API.  I will remove it from ```pywrapper.py```.

I created ```get_final_filename``` which can retrieve a filename after the ```cmor.close()``` has been called.  I used it in some of mytest.

* Test/test_python_CMIP6_CV_HISTORY.py
* Test/test_python_CMIP6_CV_furtherinfourl.py

dnadeau4:
BTW, there is not need for this since you can create your own path in the ```User Input CMOR file```.

Here is the default templates for ```CMIP6```.  These are internal and used for other projects.  If you change this, you won't be able to publish your data in ESGF.

[see template example](http://cmor.llnl.gov/mydoc_cmor3_python/)

```output_path_template":  "<activity_id><institution_id><source_id><experiment_id><variant_label><table><variable_id><grid_label><version>```
```output_file_template":  "<variable_id><table><experiment_id><source_id><variant_label><grid_label>```

ehogan:
Would it be possible for you to provide a link to the pull request / commit that contains this fix, please?

PR #85, opened 2016-08-09
Remove create_output_path from pywrapper.py

ehogan:
#67 was closed, but `create_output_path` still exists in `pywrapper.py`; this pull request removes `create_output_path` from `pywrapper.py`.


dnadeau4:
create_output_path is no longer valid.

Issue #65, opened 2016-06-16
Error handling in CMOR 3.1


ehogan:
Thank you for making the changes required to prevent the core dumps :)

I have discovered some differing behaviours in the way CMOR 3.1 handles errors.

**Example 1**

Error terminates the Python interpreter:
```
% ipython
Python 2.7.6 (default, Jan  3 2014, 16:42:21) 
Type "copyright", "credits" or "license" for more information.

IPython 2.2.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import cmor

In [2]: cmor.dataset_json('')
C Traceback:
! In function: cmor_is_setup
! called from: cmor_dataset_json
! 

!!!!!!!!!!!!!!!!!!!!!!!!!
!
! Error: You need to run cmor_setup before calling any cmor_function
!
!!!!!!!!!!!!!!!!!!!!!!!!!
%
```
**Example 2**

Error causes a `KeyboardInterrupt` exception to be thrown with no message. In addition, there is no log message, unless `cmor.close()` is called:

``` 
% ipython
Python 2.7.6 (default, Jan  3 2014, 16:42:21) 
Type "copyright", "credits" or "license" for more information.

IPython 2.2.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import cmor

In [2]: cmor.setup(logfile='cmor.log')

In [3]: cmor.dataset_json('')
---------------------------------------------------------------------------
KeyboardInterrupt                         Traceback (most recent call last)
KeyboardInterrupt: 

In [4]: more cmor.log

In [5]: cmor.close()
Out[5]: 0

In [6]: more cmor.log

C Traceback:
! In function: cmor_open_inpathFile
! called from: cmor_dataset_json
! 

!!!!!!!!!!!!!!!!!!!!!!!!!
!
! Error: Could not understand file "./" Is this a JSON CMOR table?
!
!!!!!!!!!!!!!!!!!!!!!!!!!


C Traceback:
! In function: cmor_dataset_json
! 

!!!!!!!!!!!!!!!!!!!!!!!!!
!
! Error: JSON resource file not found
!
[...]

PR #86, opened 2016-08-16
Python error handling

ehogan:
Contrary to [my previous comment](https://github.com/PCMDI/cmor/issues/65#issuecomment-227139826), it is not possible to catch exceptions using the current version of CMOR. To test whether I could catch an exception when attempting to produce an output netCDF file for a variable and then go on to produce an output netCDF file for a second (different) variable, I added the following lines to the [CMIP6 example](https://github.com/PCMDI/cmor/blob/master/Test/test_python_CMIP6_experimentID.py) after the `cmor.load_table` call:

```
    try:
        print 'In try loop'
        ivar = cmor.variable(table_entry='soga', axis_ids=[], units='0.001')
    except BaseException as err:
        print 'In except loop'
        print err
    finally:
        print 'In finally loop'
    print 'Continuing'
```

When I run the edited example I get:

```
% ./example.py
In try loop

C Traceback:
! In function: cmor_variable
! 

!!!!!!!!!!!!!!!!!!!!!!!!!
!
! Error: You are defining variable 'soga' (table Omon)  with 0 dimensions, when it should have 1
!
!!!!!!!!!!!!!!!!!!!!!!!!!

In finally loop
Traceback (most recent call last):
  File "./example.py", line 43, in <module>
    main()
  File "./example.py", line 28, in main
    ivar = cmor.variable(table_entry='soga', axis_ids=[], units='0.001')
KeyboardInterrupt
%
```

As you can see, the code doesn't get into the `except` loop (although the log message is printed at that time), and the exception is raised after the `finally` loop. This [stack overflow post](http://stackoverflow.com/questions/1717991/throwing-an-exception-from-within-a-signal-handler) says that "the signal handler runs with its own stack", which perhaps gives an indication why this is happening.

This pull request:

* raises a `CMORError` exception at the correct time
* removes the signal handler from `pywrapper.py` (it is no longer required, since the signal handling is done in `_cmormodule.c`)
* restores using a `SIGTERM` rather than a `SIGINT` (now that the signal is being handled appropriately, I don't see the core dumps any more)
* ensures the functions return the error code rather than None (I'm happy to update the documentation accordingly, if this pull request is merged)

I used the following sources when writing this code (my C is a little rusty):

* The signal handler: http://www.gnu.org/software/libc/manual/html_node/Handler-Returns.html#Handler-Returns
* Adding the new exception definition to the module's initialisation function: https://docs.python.org/2.7/extending/extending.html#intermezzo-errors-and-exceptions
* PyErr_Format: https://docs.python.org/2.7/c-api/exceptions.html#c.PyErr_Format

Running the edited example using this pull request:

```
% ./example.py
In try loop

C Traceback:
! In function: cmor_variable
! 

!!!!!!!!!!!!!!!!!!!!!!!!!
!
! Error: You are defining variable 'soga' (table Omon)  with 0 dimensions, when it should have 1
!
!!!!!!!!!!!!!!!!!!!!!!!!!

In except loop
Problem with 'cmor.variable'. Please check the logfile (if defined).
In finally loop
Continuing
! ------
! CMOR is now closed.
! ------
! During execution we encountered:
!   0 Warning(s)
!   1 Error(s)
! ------
! Please review them.
! ------
! %
```

The output netCDF file for the `masso` variable is produced successfully.

I'm afraid I am still unable to run the tests (I still get an error), but I have done some limited testing and it seems to be behaving well.

### DIFF
--- a/Lib/__init__.py
+++ b/Lib/__init__.py
@@ -1,8 +1,13 @@
 from cmor_const import *
 
-from pywrapper import axis,variable,write,setup,load_table,set_table,zfactor,close,grid,set_grid_mapping,time_varying_grid_coordinate,set_cur_dataset_attribute,get_cur_dataset_attribute,has_cur_dataset_attribute,create_output_path,set_variable_attribute,get_variable_attribute,has_variable_attribute,dataset_json,get_final_filename,set_deflate
+from pywrapper import (
+    CMORError, axis, variable, write, setup, load_table, set_table, zfactor,
+    close, grid, set_grid_mapping, time_varying_grid_coordinate, dataset_json,
+    set_cur_dataset_attribute, get_cur_dataset_attribute, create_output_path,
+    has_cur_dataset_attribute, set_variable_attribute, get_variable_attribute,
+    has_variable_attribute, get_final_filename, set_deflate)
 
 try:
-  from check_CMOR_compliant import checkCMOR
-except:
-  pass
+    from check_CMOR_compliant import checkCMOR
+except ImportError:
+    pass
--- a/Lib/pywrapper.py
+++ b/Lib/pywrapper.py
@@ -1,11 +1,9 @@
+import numpy
+import os
 
-import cmor_const,numpy,os,_cmor
-import signal
-
-def sig_handler(signum, frame):
-    os.kill(os.getpid(),signal.SIGABRT)
-
-signal.signal(signal.SIGTERM, sig_handler)
+import cmor_const
+import _cmor
+from _cmor import CMORError
 
 try:
     import cdtime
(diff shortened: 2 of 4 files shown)


ehogan:
Related to [this comment](https://github.com/PCMDI/cmor/issues/65#issuecomment-226478052), it should be easy to update `_cmormodule.c` to not raise an exception based on the error code, if necessary.

Also, as a possible future improvement, perhaps the exception message could be set based on the error code / the message that is sent to the log?

dnadeau4:
I will be looking into this next week.

ehogan:
Great, thank you @dnadeau4!

PR #96, opened 2016-09-14
Don't raise an exception for has_variable_attribute …

ehogan:
…and has_cur_dataset_attribute if the attribute doesn't exist (the functions will instead return `False`).


dnadeau4:
How come you don't want to raise an exception if one or many variable attribute(s) or a dataset attribute(s) do(es) not exist?

dnadeau4:
Can you write a small test showing what was failing when has_cur_attribute return false.  That  seems to be the correct behavior for this API.

ehogan:
@dnadeau4 I added a few tests and included them in the Travis testing. I noticed the osx build contains a `Warning! PATH is not properly set up` message and ends with `/Users/travis/build.sh: line 109: shell_session_update: command not found`; is this a problem?

ehogan:
Also, I just noticed that `cmor_set_cur_dataset_attribute()` is missing from http://cmor.llnl.gov/mydoc_cmor3_api/; added to PCMDI/cmor3_documentation#15.

dnadeau4:
Thanks Emma, this is great! :+1:

Issue #97, opened 2016-09-21
The value of the realm global attribute in netCDF file for tasmin is not equal to the value of the modeling_realm in CMIP6_day.json MIP table


ehogan:
I have produced an output netCDF file for `tasmin` using the `CMIP6_day.json` MIP table. The value for the `modeling_realm` attribute in the `CMIP6_day.json` MIP table for `tasmin` is `atmos` (https://github.com/PCMDI/cmor/blob/master/TestTables/CMIP6_day.json#L2932), but the value of the `realm` global attribute in the output netCDF file is `seaIce ocean`, which is equal to the value of the `realm` attribute in the header of the `CMIP6_day.json` MIP table (https://github.com/PCMDI/cmor/blob/master/TestTables/CMIP6_day.json#L5). Should the value for the `modeling_realm` attribute be used instead?

dnadeau4:
I talked to @taylor13 and he told me the WIP has not decided how to handle this yet.  

Sorry about the confusion, the Table header have `realm` which is a global attribute and the Variable has `modeling_realm` which is a variable attribute.   It seems as well that the `CMIP6_day` table has a mixture of `seaIce ocean` and `atmos` variables.  We might have to create 2 tables for `day` stuff, I am not sure how to handle this just yet.

durack1:
@dnadeau4 not certain what my role in this is?

dnadeau4:
@taylor13  What is the difference between realm (GLOBAL) attribute and modeling_realm (variable attribute).   Can they be so different `realm=atmos` `modeling_realm=seaIce ocean` in the same file.

dnadeau4:
Solution: used the variable `modeling_realm` as the global attribute `realm`.

dnadeau4:
Will do soon....

ehogan:
[...]

PR #167, opened 2017-05-05
Correct access to the variable-specific realm value

ehogan:
While testing CMOR 3.2.3, I discovered an issue related to accessing the variable-specific realm values from the MIP tables. This pull request fixes the issue.

Some details: `var_id` is used as an index for `cmor_vars`. However, it is not appropriate to use this index to access the appropriate variable from the MIP table (i.e. in `cmor_tables[nVarRefTblID].vars[var_id]`). 

In CMOR 3.2.2, `nVarRefTblID` was used in place of `var_id` as the index (see [here](https://github.com/PCMDI/cmor/blob/3.2.2/Src/cmor.c#L2624)). For the `Omon` and `SImon` tests I have, `nVarRefTblID` has a value of 0, meaning the realm value of the first variable in the MIP table was used. Coincidentally, the realm value of the first variable in the `Omon` and `SImon` MIP tables is the same as the realm value of the variables in my tests, so it appeared the realm was being set correctly (see #97). In CMOR 3.2.3, `var_id` is used as the index (see [here](https://github.com/PCMDI/cmor/blob/CMOR-3.2.3/Src/cmor.c#L2633)), which has a value of 4 for the tests I have. The fifth variable in the `Omon` MIP table has a realm value different to the realm value of the variables in my `Omon` tests, which is how I discovered the issue.

### DIFF
--- a/Src/cmor.c
+++ b/Src/cmor.c
@@ -2506,11 +2506,13 @@ int cmor_setGblAttr(int var_id) {
     regex_t regex;
     int numchar;
     int nVarRefTblID;
+    int ref_var_id;
     int rc;
     int ierr=0;
 
     cmor_add_traceback("cmor_setGblAttr");
     nVarRefTblID = cmor_vars[var_id].ref_table_id;
+    ref_var_id = cmor_vars[var_id].ref_var_id;
 
     if( cmor_has_cur_dataset_attribute( GLOBAL_ATT_FORCING ) == 0 ) {
         cmor_get_cur_dataset_attribute( GLOBAL_ATT_FORCING, ctmp2 );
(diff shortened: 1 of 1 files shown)


taylor13:
@ehogan @dnadeau4  Not sure I followed the details, but we need to make sure that Martin's tables provide both the CMOR table "entry" label for the variable (which the user uses to tell CMOR what table entry should be referenced) and the "output name" of the variable, which is what appears in the netCDF file as the variable name and is also the "variable_id".  Except in a few cases these are identical, but not always, I think.  (At least they weren't always the same in CMIP5.)

dnadeau4:
@taylor13 and I noticed that there is no `output_name` of the variable.  I remember when I created the SQL database I just copied `label` to `output_name`, but at the time (2 years ago already, we did not use github for CMIP6 dreq issues.)   

@taylor13 will make sure Martin knows and I will add a field to my database when it becomes available.

dnadeau4:
Looking at this issues, I **did** change `nVarRefTblID` to `var_id`.  I don't  know what `cmor_vars[var_id].ref_var_id` points to...   I will investigate.  

I might have introduced this bug trying to make it more readable after using `indent`.   

There were very very long `arrays of arrays of arrays` before....

Thanks for the PR. :+1:

dnadeau4:
Finally I found out that  `ref_var_id` points to a variable in the `cmor_table[]` array. `var_id` is just an arbitrary index for `cmor_vars[]` array.

https://github.com/PCMDI/cmor/blob/master/Src/cmor_variables.c#L1054-L1081

The `cmor_var` structure **does** have a `realm` but is not initialized using `table[].realm `. 
https://github.com/PCMDI/cmor/blob/master/Src/cmor_variables.c#L1860

**And** it can be set by the user and (then not used).....
https://github.com/PCMDI/cmor/blob/master/Src/cmor_variables.c#L2099

I will merge your PR, and then revisit this...  I think `cmor_vars[].realm` should be set and used.
@doutriaux1  Is there a reason for not using `cmor_vars[].realm`?

Thanks for your help in this issue.

ehogan:
Thanks Denis! :)

PR #171, opened 2017-05-12
Only close variables if no error was encountered

ehogan:
Prior to the inclusion of #86, if an error was encountered, CMOR would exit immediately, meaning there would be no opportunity to call `cmor_close`; this meant the name of the output netCDF file would remain with the temporary string, rather than be finalised to include the start and end dates. Now, with the inclusion of #86, errors can be caught, meaning that `cmor_close` can be called on variables where an error was encountered. This can lead to misleading results; when I see an output netCDF file with a name that contains the start and end dates, I assume the file was generated correctly. However, this assumption no longer holds.

This pull request contains an idea of how to revert the behaviour of CMOR prior to the inclusion of #86.

Some details:

* An `error` attribute was added to the `cmor_var` structure to store whether an error was encountered.
* A `cmor_handle_error_var` function was added, which accepts `var_id` (in addition to the arguments accepted by `cmor_handle_error`). `var_id` is then used to access the correct `cmor_var` and set the `error` attribute appropriately. Then `cmor_handle_error` is called as normal.
* Calls to `cmor_handle_error` were updated to instead call `cmor_handle_error_var` in cases where the `var_id` was available, but only when the level is `CRITICAL` or `NORMAL`.
* When attempting to close a variable, the `error` attribute is checked, so only variables that did not encounter an error are closed.

I've tested this and confirmed that when an error is encountered, the name of the output netCDF file is no longer finalised to include the start and end dates, but remains with the temporary string.

@dnadeau4 please would you let me know if you are happy with this idea and whether you have any suggestions for improvement? The other `*.c` files in `Src` will need to be updated in a similar fashion, but I wanted to ask for your opinion first :)

### DIFF
--- a/Src/cmor.c
+++ b/Src/cmor.c
@@ -608,6 +608,11 @@ void cmor_handle_error(char error_msg[CMOR_MAX_STRING], int level) {
     }
 }
 
+void cmor_handle_error_var(char error_msg[CMOR_MAX_STRING], int level,
+                           int var_id) {
+    cmor_vars[var_id].error = 1;
+    cmor_handle_error(error_msg, level);
+}
 
 /************************************************************************/
 /*                        cmor_reset_variable()                         */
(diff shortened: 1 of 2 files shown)


dnadeau4:
Thanks Emma for doing this work!  I might piggy back on the function as well,  I think we can change the filename in that function, or detect error and add some "_ERROR" at the end of the file.

PR #174, opened 2017-05-15
Only close variables if no error was encountered

ehogan:
@dnadeau4 I made some more changes related to #171; I continued to push to my branch, but this pull request shows the changes you already merged. I hope this won't cause any problems for you!

I'm not sure the changes to `cmor_CV.c` are required. I believe all the CV checks are run before the variable is initialised; should I revert those changes?


dnadeau4:
No problem,  github does a diff, so only new changes are merged.

Issue #178, opened 2017-05-25
time_range in filename not constructed correctly


jonseddon:
Hi,

Two issues have been spotted in the time_range segment of filenames:

* The components included are not always correct for the specified frequency, e.g.:

`ua7h_E3hrPt_HadGEM3-GC31-LM_highresSST-present_r1i1p1f1_gn_19500101030000-19501230210000.nc` (should be `yyyyMMddhhmm` rather than `yyyyMMddhhmmss`)

Other files are named correctly:
`rsdscs_3hr_HadGEM3-GC31-LM_highresSST-present_r1i1p1f1_gn_195001010130-195012302230.nc`

I think that this is due to the interval not being calculated correctly at https://github.com/PCMDI/cmor/blob/master/Src/cmor.c#L5435. In a file at a frequency of 1hr, with all of the time coordinate points correctly spaced at 3600 second intervals, interval was assigned the value of 1499.99. I don't think that this is a problem with our model data as I have seen the same problem with files generated by another institution. I haven't been able to work out why `cmor_convert_interval_to_seconds()` returns the wrong value.

* The seconds in the time_range string are always rounded down, rather than to the nearest second. Other tools that work with netCDF data round to the nearest second and so the times in the filenames don't match the times reported by other tools, e.g.:
`prc_E1hr_HadGEM3-GC31-LM_highresSST-present_r1i1p1f1_gn_19500101002959-19501230233000.nc`

where the first time is 1950-01-01 00:29:59.99993.

I've got a fix for these and will raise a PR shortly.

Thanks,

Jon

durack1:
@jonseddon thanks for this.. If there was a way to attempt to ensure that most of the MOHC contributed data starts in January and ends in December for the contributed years that would make life considerably easier for a user compared to CMIP5.. Just a user comment, unrelated to this issue

taylor13:
@dnadeau4 @jonseddon :  Just to confirm that the precision required by CMIP6 for different frequencies can be found in table 2 of https://docs.google.com/document/d/1h0r8RZr_f3-8egBMMh7aqLwy3snpD6_MrDz1q8n5XUk .  For 3hr data, we require yyyyMMddhhmm (no seconds), as in the 2nd example above.

dnadeau4:
@jonseddon WOW!  This is awesome,  thanks for the code!   I'll run my tests with it and if filenames are fine, then I will merge.   (I'll let @doutriaux1 and @taylor13 check before)

[...]

PR #179, opened 2017-05-25
Round seconds

jonseddon:
This is for #178.

As interval is not always returned correctly by `cmor_convert_interval_to_seconds()` the frequency is taken from the metadata instead and the time_range segment of the filename is constructed from that.

Seconds are rounded to the nearest whole second before minutes or seconds are included in the time_range of the filenames.

All tests pass. This is my first look at CMOR and I've not written any C start I started using Python and so I apologise if I've made any style or other faults in my code!

Thanks,

Jon


jonseddon:
Resolved by #178

Issue #246, opened 2017-10-04
Differences between variable name and out_name


ehogan:
In the `01.00.15` version of the MIP tables, there are a number of variables that have an `out_name` that differs from the variable name, but the variable `out_name` already exists in the MIP table. See, for example, [ua in CMIP6_6hrPlevPt.json](https://github.com/PCMDI/cmip6-cmor-tables/blob/master/Tables/CMIP6_6hrPlevPt.json#L484) and [ua7h in CMIP6_6hrPlevPt.json](https://github.com/PCMDI/cmip6-cmor-tables/blob/master/Tables/CMIP6_6hrPlevPt.json#L502), which has an `out_name` of `ua`.

This means, once `cmor.setup` and `cmor.dataset_json` have been run, if someone tries to produce `ua` from `6hrPlevPt`, then `ua7h` from `6hrPlevPt`, wouldn't the corresponding filename of the output netCDF file be the same? Wouldn't this cause issues?

ehogan:
cc @martinjuckes @taylor13 @dnadeau4

dnadeau4:
Yes the filename will be exactly the same and one file will be clobbered. 
My guess is that data provider will select which levels the want
https://github.com/PCMDI/cmip6-cmor-tables/blob/master/Tables/CMIP6_coordinate.json#L1603
https://github.com/PCMDI/cmip6-cmor-tables/blob/master/Tables/CMIP6_coordinate.json#L1648

durack1:
@ehogan will there be a real world example that both of these variables (with different height axes) will be written at the same time?

@taylor13 if the answer to the above is yes, then we need to determine how to deal with this, as both files will share the same filename and DRS (they originate from the same table)

ehogan:
@durack1 I don't currently have a real world example that demostrates this issue. Perhaps a better question would have been "What is the reasoning behind using an `out_name` that is different to the variable name?"

taylor13:
The table "entry" (name) can distinguish between two variables appearing in the CMOR table that have different attribute values.  Since the requested pressure levels are different between ua and ua7h, they have different table "entries".  For a given experiment, they shouldn't both be requested.  In particular if ua7h is needed that will already include ua (which only has 3 pressure levels requested (assuming the 3 are included in the 7 levels of ua7h).  So there is no danger that two files will be produced with the same name.
[...]

PR #249, opened 2017-10-09
Use the final name of the file (rather than the temporary name) when …

ehogan:
… renaming so that subsequent files with the same name as the final name can be written.

See #246.

I also moved the call to `cmor_handle_error_var` to the case where the original file can not be renamed; if the original file was successfully renamed, CMOR can continue as normal.

### DIFF
--- a/Src/cmor.c
+++ b/Src/cmor.c
@@ -5965,7 +5965,7 @@ int cmor_close_variable(int var_id, char *file_name, int *preserve)
             fperr = fopen(outname, "r");
             if (fperr != NULL) {
                 sprintf(msg, "%s.copy", outname);
-                if (rename(cmor_vars[var_id].current_path, msg) == 0) {
+                if (rename(outname, msg) == 0) {
                     snprintf(msg, CMOR_MAX_STRING,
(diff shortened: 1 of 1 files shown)


taylor13:
As I wrote at https://github.com/PCMDI/cmor/issues/246 ...

I think you are trying to do something that we don't want you to do: write both ua and ua7h. ua is redundant because ua7h already includes the ua field. Therefore, to avoid the error, simply stop trying to write "ua' when you've already written (or are planning to write) ua7h. Then you won't encounter the problem. I don't think there is any reason to rename any files produced by CMOR. Do you disagree?

dnadeau4:
@ehogan is this still valid?  I did not merge since @taylor13 made a comment on the `ua7h` and `ua` variable.

ehogan:
@dnadeau4, if @taylor13 can comment on [my comment](https://github.com/PCMDI/cmor/issues/246#issuecomment-336455017) from Oct 13, 2017 on #246, we can make a decision about this pull request :)

dnadeau4:
@ehogan could you write a test for this?  just to make sure it works the way you want and we understand the changes.  Thanks!

PR #261, opened 2017-10-26
User/aradhakrishnan gfdl/check axis fix

aradhakrishnanGFDL:
This fix aims at identifying scalar singleton axis using value != "" in axis definition. Previously tested on a tas failure case, ts success case. 

New test on a succes tas test-case:  float tas(time, lat, lon) ;
('Warning: There might be a singleton dimension associated with this ', u'height', ". Please make sure it's present in the coordinates attribute")
Dimension check:OK

Please note that the test was performed in a local python test environment at GFDL only. Thanks.

### DIFF
--- a/LibCV/PrePARE/PrePARE.py
+++ b/LibCV/PrePARE/PrePARE.py
@@ -23,7 +23,6 @@ import sys
 import os
 import json
 import numpy
-from pprint import pprint
 
 class bcolors:
     HEADER = '\033[95m'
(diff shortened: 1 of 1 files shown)


dnadeau4:
@aradhakrishnanGFDL this won't work for us, did @taylor13 explained why?  We have `singleton` dimensions that are added for some variables.  For example, `ts` has `height2m` which is added automatically.  This would be difficult for PrePARE to handle an extra dimension.

dnadeau4:
I will close this PR since we won't be able to implement this.

PR #274, opened 2017-12-13
PrePARE

dnadeau4:


durack1:
@dnadeau4 did you want to release `3.3.0` or `3.2.9` which is what it appears is included here

PR #285, opened 2018-01-08
Add multiprocessing to PrePARE

glevava:
- Handle multiprocessing with redirected output in threads logfile
- Deduce cmip6 table from filename


dnadeau4:
@glevava I merged your PR into my PrePARE branch.  I will merge it to master if the tests passes.
Thanks for your help with parallelization.  I modified the input fields, but all the functionality stay the same.  I need to look at the log a little bit more carefully as well when I get more time.

PR #292, opened 2018-01-25
Add necessary changes for VPATH build.

tjahns:
Hello,
I've put in the necessary changes so cmor can be built from another directory like automake projects. As a side-effect, cross-compilation also should be smoother.
Regards, Thomas

* With these modifications it becomes possible to execute
  configure in another directory and have mulitple builds
  based on the same source tree.
* Also since show_git is no longer needed, cross-compilation
  becomes much easier.


dnadeau4:
Thanks for this!  I just pushed a branch and if `travisci` passes, I will merge it.

dnadeau4:
merge using a new PR which triggered `travis-ci`

PR #304, opened 2018-02-20
Cleanup code and arguments. 

glevava:
"cmip6_table" positional input has been removed to avoid conflicts with multiprocessing.
PrePARE only deduce CMOR table from the filename. No need to submit the appropriate table.

### DIFF
--- a/LibCV/PrePARE/PrePARE.py
+++ b/LibCV/PrePARE/PrePARE.py
@@ -139,6 +139,7 @@ class Spinner:
         sys.stdout.flush()
         Spinner.step += 1
 
+
 # =========================
 # checkCMIP6()
 # =========================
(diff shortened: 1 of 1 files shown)


taylor13:
If I understand this correctly, a cmor table name must be obtainable from the file name.  There are projects (other than CMIP) where CMOR will be used and the table name won't be in the file name.  We need to check that before merging.  If there is a problem, would CMOR/PrePARE work as it is if you limited its use in parallel to variables found in a single table?  Would that be o.k.?

glevava:
Yes the cmor table can be obtainable from the filename. But I just realize that isn't necessarily the case for all project. So you're right, we should support initial behavior for other project than CMIP or CORDEX. Feel free to rejet the PR and I can adapt the code to support both behaviors with different flags, for instance `--table-path` and `--table-dir` in a mutually exclusive group.

PR #307, opened 2018-02-21
PrePARE call review

glevava:
- `--table-path` can be a full JSON path or the table directory. If a directory, the CMOR table is deduced from filename
- `--variable` is optional. If not submitted, the variable is deduced from filename
- `--table-path` and `--variable` cab be used in sequential or multiprocessing mode. With multiprocessing, pay attention to scan files corresponding to the same CMOR table and/or variable if the flags are used.
- bugfix on logfile in multiprocessing mode: now print statements are "append" to logfile instead of truncate the logfile (all file diagnostics are now returned at the end of the multiprocessing file checkup).
- code has been simplified, reformatted and cleaned.

### DIFF
--- a/LibCV/PrePARE/PrePARE.py
+++ b/LibCV/PrePARE/PrePARE.py
@@ -26,8 +26,11 @@ sys.path.insert(0, sys.prefix + "/lib/python2.7/site-packages/cdms2")
 import Cdunif
 import argparse
 import os
+import itertools
+from argparse import ArgumentTypeError
 import json
 import numpy
+from uuid import uuid4 as uuid
 import cmip6_cv
 from multiprocessing import Pool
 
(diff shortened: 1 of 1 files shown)


glevava:
Hi @taylor13 , this PR cancels and replaces the previous one. I slightly modify the code to allow JSON file path and table directory as input for PrePARE.

taylor13:
Sounds like a nice contribution, but I'll leave it to the expert (@dnadeau4 ) to determine if it can be merged into the master.
Thanks!

glevava:
@taylor13 @dnadeau4 Of course ! No hurry as I also pull a request for the ESGF publisher to directly call PrePARE methods by sourcing the code instead of faking the arguments. Thus it is not  affected by the issue. I also receive a recommendation from CNRM to add some "filters" on the input directory to include or exclude file/directory patterns from the PrePARE check (when use as standalone outside of the ESGF publisher). Maybe for another PR ; ) Thanks !

durack1:
@glevava did you mean to close this?

glevava:
Yes, closed and replaced by https://github.com/PCMDI/cmor/pull/315

PR #322, opened 2018-03-15
Add safe return code

glevava:
This rollback different return code depending on PrePARE diagnostic:
-   0 = no errors
-   more than 0 = errors (with code = number of files with error(s)) 
-   -1 = all files has error(s)

Note: this doesn't work properly as `cmip6_cv.check_filename()` doesn't return 1 in case of wrong filename. See [https://github.com/PCMDI/cmor/issues/323](https://github.com/PCMDI/cmor/issues/323)


dnadeau4:
I like this.  I working somewhere else this week  and will get to it next week.

PR #353, opened 2018-05-23
remove env section, using travis web to set env variables

doutriaux1:


durack1:
@doutriaux1 I'm using CMOR on mac too, so would be great to provide both versions through anaconda

doutriaux1:
@durack1 no worries for simplicity the "branch" test was only for Linux, since it works it will now upload for both mac and linux.

durack1:
@doutriaux1 perfect!

PR #358, opened 2018-06-06
PrePARE glevava

dnadeau4:


dnadeau4:
@glevava I merged the **new climatology** code with yours.  
I have tested your code and everything works.

BTW,  just be careful,  when you add an option "--all" which changes the behavior of PrePARE a lot of users out there will complain that the program does not work since they don't see "SUCCESS" by default.   As well, you changed "--max_threads" to "--max_processes".  People scripts will fail and they will write an issue.   

When we release we have to make sure that all theses changes are documented.

Thanks a lot for the PR and the clean up.

dnadeau4:
@glevava I see that you select a `--max-processes` to 4 by default.  I tripped on this since I could not longer use the debugger, until I read your `synopsis`.   We might want to tell `ESGF` about this.
I like the `-1` option. 👍

dnadeau4:
@taylor13 @doutriaux1 this PR can be merged.

glevava:
Hi @dnadeau4 I didn't think my latest commits was propagated to the old PR. I thought to make a new PR to discuss with you of course, especially regarding the behavior changes with --all. I implemented this option because at IPSL we check thousands of fils in a bulk and PrePARE can be very verbose with the "SUCCESS" statement that I don't care because obviously ; ) 
Regarding --max-process default to 4 it doesn't affect ESGF as all other changes in the arguments because the esg-publisher directly import and call the check_cmip6 validator from PrePARE code. It doesn't care about the multiprocessing. Maybe the "--all" behavior can affect it as the "SUCCESS" statement are now outside of the validator... Need to check this...
But basically those latest changes can be part of a personnal PrePARE version at IPSL, I can make another branch for us.

Thanks for all !

glevava:
@dnadeau4 I made a small change in the print statements to keep publisher behavior as before (i.e., print all statements and not only errors which the default in my branch when using the command-line). See my latest commit https://github.com/glevava/cmor/commit/b12a0955c85a6c0370e4c02ba0212ce50b3da6df 
If it's fine with you I can make a new pull request :/ ...

dnadeau4:
@glevava I can use this PR and merge your change.   
Can you check this code and tell me if it is still needed?  I think your new functions now take cat of this and this can be deleted.

https://github.com/PCMDI/cmor/pull/358/commits/83c9f591656cd10917451bf6fca3693718f4522b

dnadeau4:
https://github.com/PCMDI/cmor/blob/83c9f591656cd10917451bf6fca3693718f4522b/LibCV/PrePARE/PrePARE.py#L471-L479

glevava:
Yes the new functions include the "climatology" case. See in the whole file of my pull request this piece of code has been delete and the block to get time_bounds also.

glevava:
FYI, I use my PrePARE.py file since several weeks now at IPSL site to check all of our CMIP6 simulation in PROD. We never find any "technical" bug from my PrePARE version. If it can help you to do the merge.

dnadeau4:
@glevava Thanks,  I still need to rerun all the test and I now work only 1 day/month for PCMDI.  It is hard to merge your file since you move things around.  What happens is the merge add your new code and I have duplicated code.   I don't always catch the duplication.

glevava:
Hi @dnadeau4 I would say you can take my PrePARE code as it instead of dealing with merge diff.
But I don't know how my PrePARE modifications complies with your test suite, I 'm sorry for that.

dnadeau4:
@glevava I merged the PR and deleted the branch.

This is how we run our tests.   

https://circleci.com/gh/PCMDI/cmor/152?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Here is the configuration:
https://circleci.com/gh/PCMDI/cmor/152#config/containers/0

PR #380, opened 2018-07-30
01.00.27

doutriaux1:
@taylor13 @durack1 please review changes brought in by 01.00.27


durack1:
@doutriaux1 these changes from a quick glance look good to me, it's just catching up on the standard and long_name changes that were proposed quite sometime ago. There are also some updates to the coordinates `gridlatitude` -> `grid_latitude` etc. Looks good to me, @taylor13 may have some more insight, so I suggest we await his review

doutriaux1:
@durack1 I will review the grid_latitude issue, you're right, there was some noise about it last week.

durack1:
@doutriaux1 thanks, that's the only thing (aside from very long comments or standard names) that I saw that could cause software to fail

TestTables/CMIP6_coordinate.json
```diff
@@ -1290,7 +1268,7 @@
             "type": "double", 
             "valid_max": "", 
             "valid_min": "", 
-            "value": "84000.", 
+            "value": "m",
```
doutriaux1: @taylor13 is that right?

TestTables/CMIP6_coordinate.json
```diff
@@ -608,7 +585,7 @@
             "bounds_values": "", 
             "generic_level_name": ""
         }, 
-        "grid_latitude": {
+        "gridlatitude": {
```
doutriaux1: @taylor13 is this expected?

TestTables/CMIP6_coordinate.json
```diff
@@ -1,28 +1,5 @@
 {
     "axis_entry": {
```
doutriaux1: @taylor13 do we really want to yank this?

doutriaux1:
@durack1 @taylor13 I just addressed lat/lon std name and long name issue. Please review the new string are adequate.

TestTables/CMIP6_coordinate.json
doutriaux1: @taylor13 interestigly this does not show up on cmip6-tables changes because it appears to have been introduce in26. The actual cmor tables in the code are still at 24, which is why we notice it here but in the cmip6-cmor-tables.

taylor13: This is definitely an error in the table.  value *should* be 84000.  Can you edit the CMOR dimension table by hand before committing?  I'll inform Martin it needs to be corrected for release 28.

taylor13:
Do you want me to review something here?

doutriaux1:
It's merged now. You did review it by email.

PR #387, opened 2018-08-03
385 terminate signals

doutriaux1:


doutriaux1:
@ehogan @taylor13 this should address the sigint/sigterm issue

doutriaux1:
@mauzey1 might be a good easy one for you review and take a look at
kinds of cover a bunch of things in CMOR bindings with other languages

doutriaux1:
The idea is that C and Python revert to SIGTERM by default and Fortran stays with SIGINT for default

While at it I add a few function to let the user set the terminate signal to whatever signal is desired.

PR #398, opened 2018-08-23
Added test for appending associated variables

wachsylon:
E.g. ps for hybrid vertical axes


wachsylon:
There seems to be a bigger problem with the append mode for this case. All time values after second time step are incorrect in the outfile when writing in append mode - even when writing the first file without suitable chunk file to append on.

An explanation could be that I made a mistake in that test - but I can not see any. Could you check?

PR #404, opened 2018-09-19
Cmor append time error chris

mauzey1:


wachsylon:
Thank you very much for improving! I actually do not pass time values in `cmor_axis` in my program but lost the clear view while writing the test, sorry for that. There is one difference: I also pass time values in `cmor_write` in the first step and not in `cmor_axis`. But that should not make a difference. However, I will test it.

wachsylon:
OK this has also worked!

PR #419, opened 2018-12-13
first stab at yanking files

doutriaux1:


durack1:
@doutriaux1 thanks for this, it'll be great to have a cleaner repo moving forward - and also good so not to confused users who don't read the documentation

Issue #430, opened 2019-01-15
singleton and generic level


doutriaux1:
Koji OGOCHI <ogochi@jamstec.go.jp>
Japan Agency for Marine-Earth Science and Technology (JAMSTEC)

reports:

```
I found a serious problem that appears if a variable
has both generic_level and singleton in its dimensions.
The variable ec550aer is the case.

I fixed this problem like the attached patch.

Regards,
Koji Ogochi
```

Patch attached.

[cmor_variable.patch.txt](https://github.com/PCMDI/cmor/files/2760124/cmor_variable.patch.txt)

taylor13:
This variable is in CMOR table "Emon" where ec550aer has coordinates:
longitude latitude alevel time lambda550nm
(and lambda550nm is a singleton coordinate).  

The text file linked to above seems to be corrupted.  Is there a clean version?

martinjuckes:
Just a comment: ec550aer (Emon and 6hrLev) appear to be the only variables which have both a vertical dimension and singleton coordinate variable. This coordinate variable is a wavelength, so it should be acceptable as far as CF is concerned.

taylor13:
Yes, I think the variable is defined correctly and consistent with the CF conventions.  We'll need to check that CMOR, with the patch, handles everything correctly.

taylor13:
See also https://github.com/PCMDI/cmip6-cmor-tables/issues/230

wachsylon:
I get the same Error for `rsdcsafbnd` in `E3hrPt`.

taylor13:
It's not clear that the two errors are really the same.   rsdcsafbnd in E3hrPt is a function of 5 coordinates: longitude latitude alevhalf spectband time1.  Have we tested whether CMOR correctly deals with 5?  Can we generate a test case?

Regarding ec550aer, it is also a function of 5 coordinates, but the last one is a scalar dimension (not a true coordinate), so it gets handled differently by CMOR.  This would require a different test code. 

We should try to fix these bugs as soon as we can.

taylor13:
Looking at the patch, I'm concerned that it might break something else (not tested by our test codes).  Do we understand what Ogochi has done?

mauzey1:
@taylor13 
The changes in Ogochi's patch are fixing a bug in cmor_variable() where it's getting the 'Z axis' dimension for the lists of axes in a variable.
[...]

PR #431, opened 2019-01-15
Patch for issue #430

mauzey1:
Tested using this file

https://github.com/mauzey1/cmor/blob/daaefaa504609e3c6536d6e2a0ff923a8e1efcd0/Test/test_cmor_python_singleton_and_generic.py


taylor13:
I don't think this patch should be merged, even if it passes current tests, unless the logic of what is being done is completely understood, and we can be reasonably certain that making this change won't break something that is not currently checked in our test suite.  I would be happy to go over the coding with someone who knows C better than I do.

PR #455, opened 2019-03-22
Fixed if statement in conda_upload.sh

mauzey1:


taylor13:
Before making this release, can we make the changes called for in https://github.com/PCMDI/cmor/issues/457 ?

PR #467, opened 2019-04-03
Check for string types using six.string_types

### LINKED ISSUE #466: isinstance(var, str) breaks unicode
https://github.com/PCMDI/cmor/blob/11f7665d23e0780472e6d9d182a084d8b08f3ad6/Lib/pywrapper.py#L66

The checks for str, while it seams like a good idea, actually breaks any python 2.7 that uses unicode_literals. All these checks should be removed, or changed to make sure normal ASCII strings work too.

mauzey1:
Replace `str` with `six.string_types` to allow unicode_literals when checking if a variable is a string in Python 2.7.

Fixes #466


sterlingbaldwin:
Thanks for the quick turn around on my issue!

mauzey1:
@sterlingbaldwin @dnadeau4 Could either of you provide a review approving this PR?

doutriaux1:
@mauzey1 if possible adding a test would be nice. Also we need to add `six` to `meta.yaml.in`

mauzey1:
@doutriaux1 I added a test for handling unicode strings to the Python CMIP6 CV tests.  Are we ready to merge this branch?

Issue #464, opened 2019-04-01
Two or more singleton dimensions (bs550aer)


kjoti:
Hi, i am koji.

The problem occurs in the following cases:

1. The variable has two or more singleton dimensions
2. Users omit the 'axes_ids' of these singleton dimensions when invoking cmor_variable() 

CMOR automatically adds the singleton dimensions if needed,
so omitting them is allowed and very common.

It's easy to fix this problem. Apply the following patch:

    diff --git a/Src/cmor_variables.c b/Src/cmor_variables.c
    index 24ca3ff..dc67faa 100644
    --- a/Src/cmor_variables.c
    +++ b/Src/cmor_variables.c
    @@ -1533,7 +1533,7 @@ int cmor_variable(int *var_id, char *name, char *units, int ndims,
                                       1, &pAxis->value, 'd', NULL, 0, "");
                         }
     
    -                    laxes_ids[olndims] = k;
    +                    laxes_ids[olndims + lndims] = k;
                         lndims += 1;
                     }
 
'olndims' does not change while looping and 'lndims' is incremented
(starting with zero). So in the case of only one singleton dimension,
the previous code is ok.
But generally the index of 'laxes_ids' should be 'olndims + lndims'.

mauzey1:
@kjoti 
Hi Koji.  Thank you for helping solve this problem.

Could you submit these changes as a pull request to our repo's master branch?  You can do this by making a fork of this repo, making changes in your fork, and submitting a pull request to merge your fork's branch into ours.  This makes it easier to track changes provided by contributors to this project.

Thank you,

Chris

kjoti:
@mauzey1 
Please check #468

PR #468, opened 2019-04-04
Fix an error about bs550aer

kjoti:
This error occurs if a variable has two or more singleton
dimensions and these axis ids (axes_ids) are omitted
when invoking cmor_variable().


doutriaux1:
@kjoti please add a test case. @mauzey1 please review as well.

kjoti:
@mauzey1 
It's mysterious.

There are three tests in `Test/test_singletons.c`. Even when removing `+ lndims`, the test "2 of 3" and "3 of 3" succeed, but "1 of 3" should fail.

Please check output files for each test if your test execution terminates normally. In my execution, a segmentation fault has occurred in "1 of 3" if `+ lndims` is removed.

Test/test_singletons.c
```diff
@@ -0,0 +1,163 @@
+/*
+ * test_singletons.c -- test singleton (scalar) dimensions.
+ */
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+
```
mauzey1:
```suggestion
#include <string.h>

```

Test/test_singletons.c
```diff
@@ -0,0 +1,163 @@
...
+
+int test_bs550aer(const int *axes_ids, int num_axes, double basetime)
+{
+    double time, time_bnds[2];
+    int var_id;
+    float values[8], miss = 1.e20f;
+    char positive = '\0';
+    int i;
+
```
mauzey1:
```suggestion
    int scatangle_found = 0;
    int wavelength_found = 0;

```

Test/test_singletons.c
```diff
@@ -0,0 +1,163 @@
...
+    if (cmor_variable(&var_id, "bs550aer", "m-1 sr-1",
+                      num_axes, (int *)axes_ids, 'f', &miss,
+                      NULL, &positive, NULL, NULL, NULL) != 0)
+        fail("cmor_variable(bs550aer)");
+
```
mauzey1:
```suggestion

    // Find singleton dimensions for bs550aer
    for(i = 0; i < cmor_vars[var_id].ndims; ++i){
        if(strcmp(cmor_axes[cmor_vars[var_id].singleton_ids[i]].id, "scatangle") == 0)
            scatangle_found++;
        if(strcmp(cmor_axes[cmor_vars[var_id].singleton_ids[i]].id, "wavelength") == 0)
            wavelength_found++;
    }
    if(scatangle_found != 1 || wavelength_found != 1)
        fail("error in singleton dimensions");

```

mauzey1:
@kjoti I have added some suggested changes to the test.  The changes will allow the test to see if the singleton dimensions were added to the variable.  You can commit all three suggestions as a batch.

kjoti:
@mauzey1 
Thank you for your suggestion.

mauzey1:
Fixed #464

wachsylon:
Hi @mauzey1,
for me the 
Test/test_singletons.c 
is the only test that fails for the 3.6.0 release when I use the 6.9.32 Tables. It says:

`! Error: You are defining variable 'bs550aer' (table 6hrLev)  with 3 dimensions, when it should have 5`

Is there something that I need to change?
Best regards,
Fabi

mauzey1:
@wachsylon In the 6.9.32 tables, we [removed the `scatter180` dimension from the variable `bs550aer`](https://github.com/PCMDI/cmip6-cmor-tables/commit/540cf81958aa6db5481234ac6f0839e898ea8db8).  We then changed [test_singleton.c](https://github.com/PCMDI/cmor/commit/875abc07650fcddcca968066e50b662c7c585d21) to reflect this change.

I have tested test_singleton.c using the latest build of CMOR 3.6.0 and the 6.9.32 tables and did not get the error.  Check if your test file is the latest from the repo, and make sure that it is using the directory with the 6.9.32 table.

wachsylon:
Thanks, it works with the recent test.

Maybe it is possible to design these tests independently from the tables. The failing test suggests that the cmor version is incompatible with the table version which is probably not true.

Another idea would be to tag a cmor 3.6.1. But I know this is a lot of effort for one test...

kjoti:
This test was for multiple singleton axes. `bs550aer` was the only case of having multiple singleton axes (`lambda550nm` and `scatter180`).

Since `scatter180` has been removed from `bs550aer`, this test becomes useless and can be discarded.

Issue #477, opened 2019-05-06
Append mode for instantaneous variables


wachsylon:
Hi,
when I try to append data to a chunk of a variable with `cell_methods = "time: point"`, I receive:
```
C Traceback:
In function: cmor_validateFilename
! 

!!!!!!!!!!!!!!!!!!!!!!!!!
!
! Warning: NetCDF Error (-49: NetCDF: Variable not found) looking for time bounds
! of variable 'mrsol' in file: /work/bm0021/k204210/MPIESMTEST/cmip/archive//CMIP6/ScenarioMIP/MPI-M/MPI-ESM1-2-HR/ssp245/r1i1p1f1/6hrPlevPt/mrsol/gn/v20190506/mrsol_6hrPlevPt_MPI-ESM1-2-HR_ssp245_r1i1p1f1_gnDLFthZ26639.nc
!
!!!!!!!!!!!!!!!!!!!!!!!!!


C Traceback:
! In function: cmor_write
! 

!!!!!!!!!!!!!!!!!!!!!!!!!
!
! Error: time_vals and time_bounds must be passed through cmor_write when in append mode
!
!!!!!!!!!!!!!!!!!!!!!!!!!


```

Could we enable CMOR to append data of instantaneous variables?
Best regards,
Fabi

taylor13:
@mauzey1 I think fixing this bug should be a high priority.

taylor13:
@wachsylon  @mauzey1 As a note, I think this may be a false identification of the problem.  I can't believe no one has tried to append data for instantaneous variables with CMOR3.  There's lots of data of this kind in the archive.  Can the error be explained by anything else?

mauzey1:
@taylor13 With the merging of #480, should we close this issue?

taylor13:
Yes, thanks.  I'll close it.  If @wachsylon thinks the problem hasn't been solved by https://github.com/PCMDI/cmor/pull/480 , then he can reopen.

PR #480, opened 2019-05-07
Append point

wachsylon:
Hi,
that pull request is related to #477 
I added the condition that the time dimension of the variable has to have the attribute `must_have_time_bounds` at two places for checking time bounds in `append_mode`.

In line 4369 the time axis is expected to be listed as index 0 of axes_ids. Can you check whether that is always true in append_mode?

Probably a new test should be developed checking whether the difference between the time value of the data to be appended and the time value of the chunk is as big as the frequency allows.

I also copied the `test_cmor_append_associated.c` to `test_cmor_append_associated_point.c ` where I changed the CMOR Variable to `ta` in `6hrLev` so that an instantaneous variable is checked.

Best regards,
Fabi


taylor13:
regarding "the time axis is expected to be listed as index 0 of axes_ids", I know this: the netCDF library requires the time dimension be the most slowly incremented of all the true dimensions of a variable, which would make it dimension 0 (in C convention).  In CMOR1 there was a distinction between a list of axes associated with a variable, which included any "scalar" axes (like "height" in the tas variable), and the actual dimensions defined in the netCDF file, which omits scalar axes and stores them instead as separate "coordinate" dimensions.  If this is still true in CMOR3, then time will always be index 0 when referred to as a netCDF dimension, but it might not be 0 in the list of axes (which includes scalar dimensions).

taylor13:
I think the test code should try to process variable `ec550aer` in table 6hrLev, rather than `ta`.  That way we can check whether the scalar dimension, `lambda550nm` (which in the table is listed as the 1st dimension, but it doesn't get written as an explicit dimension), is handled correctly.

wachsylon:
> Probably a new test should be developed checking whether the difference between the time value of the data to be appended and the time value of the chunk is as big as the frequency allows.

I wanted to say that this might need to be checked in the code. What happens if someone tries to append data with an earlier time step to a future chunk?

mauzey1:
@taylor13 @wachsylon 
Did we want to change the test to use the variable `ec550aer`?  Are there other tests that should be done?

ehogan:
> I think the test code should try to process variable `ec550aer` in table 6hrLev, rather than `ta`. That way we can check whether the scalar dimension, `lambda550nm` (which in the table is listed as the 1st dimension, but it doesn't get written as an explicit dimension), is handled correctly.

I'm having problems producing `ec550aer` for `Emon`; the scalar dimension `lambda550nm` isn't handled correctly:

```
!!!!!!!!!!!!!!!!!!!!!!!!!
!
! Error: Could not find all axes for this variable. 
! Check that all your axes name matches the ones found in 
! the table Emon. 
!
!
!!!!!!!!!!!!!!!!!!!!!!!!!
```

Normally CMOR deals with scalar dimensions automatically.

mauzey1:
@ehogan Are you using the stable release of CMOR 3.4.0 from conda-forge or the nightly release?

I believe the issue with variable `ec550aer` and dimension `lambda550nm` was solved in https://github.com/PCMDI/cmor/issues/430 with https://github.com/PCMDI/cmor/commit/5ffabe785d37f5d0c07aa2f8d0c62c3f7c07df27.  These changes will be part of the next stable release of CMOR (3.5.0), which we plan to release this month.

ehogan:
Thanks @mauzey1; we're using CMOR v3.4.0; I'll try again once CMOR v3.5.0 is released :)

PR #483, opened 2019-05-15
New checks for cell_measures in PrePARE

### LINKED ISSUE #439: PrePARE  cell_measures changed in 3.4.0
I have a file with "cell_measures": "area: areacello". The MIP table for this variable (Omon.ficeberg) has "cell_measures": "area: areacello volume: volcello".

PrePARE from CMOR 3.3.2 failed this file. PrePARE from CMOR 3.4.0 is passing this file. Both versions were installed from Conda. Which behaviour is correct? We were expecting a failure. 

I don't know if I'm reading it incorrectly, but from https://github.com/PCMDI/cmor/blame/master/LibCV/PrePARE/PrePARE.py#L586 I think that the regexp only returns a value if the cell_measures in the table contains "OR" (for example "area: areacello OR areacella" in SImon.sisnmass in 01.00.23 of the tables). I think that if there isn't an OR then values will be an empty list and the rest of the checks for cell_measures will pass.

The netCDF file that I've been checking is available at https://drive.google.com/file/d/1_O8JukXI6bhdIhNRPyxvWsroWsFsyQqb/view?usp=sharing

Thanks,

Jon

mauzey1:
Resolves #439


taylor13:
I couldn't find the error message that the user will receive under the following condition:
```
 PrePARE should raise an error if table value is "" but a cell_measure is defined in the file.
```
We don't want files where `cell_measures = ""` .  The attribute should not be found in the file.  Could you please check this?

mauzey1:
@taylor13 I discovered that the loop through the table attributes will skip attributes that have empty stings in the table.

Should we go through the list of attributes of a variable in a file to make sure they were defined in the table?  Or just do this for cell_measures?

taylor13:
Not sure.  Does it go throught **all** of the entries in the table or only the "attributes".  The attributes are, for example, 
```
           "frequency": "day", 
            "modeling_realm": "atmos", 
            "standard_name": "specific_humidity", 
            "units": "1", 
            "cell_methods": "area: time: mean", 
            "cell_measures": "area: areacella", 
            "long_name": "Specific Humidity", 
             "comment": "Specific humidity is the mass fraction of water vapor in (moist) air.", 
```
Does it check only some of these?

mauzey1:
https://github.com/PCMDI/cmor/blob/20f09625279b2edb0694f1c107164450e2e67951/LibCV/PrePARE/PrePARE.py#L561
The function `cmip6_cv.list_variable_attributes(varid)` will return the following for `zhalfo` in `CMIP6_Omon.json`

```
comment: Depth below geoid
_FillValue: 1.00000002004e+20
long_name: Depth Below Geoid of Interfaces Between Ocean Layers
standard_name: depth_below_geoid
cell_methods: area: mean where sea time: mean
cell_measures: area: areacello
units: m
missing_value: 1.00000002004e+20
```
`frequency` and `modeling_realm` are defined but they don't show up.  If I were to make one of those attributes an empty string in the table, then `list_variable_attributes` won't return it.

LibCV/PrePARE/PrePARE.py
```diff
@@ -634,6 +633,15 @@ def ControlVocab(self, ncfile, variable=None, print_all=True):
                 print("=====================================================================================")
                 print(BCOLORS.ENDC)
                 self.errors += 1
+        # Check if cell_measures is defined in the file but not in the table
```
mauzey1: @taylor13 I added this new section to check if cell_measures got defined in the file but not in the table.

mauzey1:
@taylor13 Do you approve of the test for cell_measures?  If PrePARE finds the attribute cell_measures in a file but sees that it is not defined in the variable table, then it will raise an error.

PR #489, opened 2019-05-22
Add libnetcdf 4.6.2 to the run requirements of meta.yaml.in

mauzey1:
I forgot to add libnetcdf 4.6.2 to the run requirements of meta.yaml.in.  This prevented the nightly builds from being uploaded.


durack1:
@mauzey1 it would be nice to add ourselves another issue, so when conda-forge cleans up the mess, we can return to build against the **latest** version of libnetcdf rather than being pinned to 4.6.2 for eternity

mauzey1:
Since we are replacing cdms2 in #493, I will close this PR.

PR #493, opened 2019-05-24
Replace cdms2 with netCDF4 in PrePARE and Python tests

mauzey1:
Resolves #490 

This PR replaces the use of cdms2 in PrePARE and some of the Python tests with netCDF4.

Note: The MacOS CircleCI checks failed due to an error when creating the conda environment.

```
MD5MismatchError: Conda detected a mismatch between the expected content and downloaded content
for url 'https://repo.anaconda.com/pkgs/main/osx-64/libcxx-4.0.1-h579ed51_0.tar.bz2'.
  download saved to: /Users/distiller/project/workspace/test_macos_cmor/miniconda/pkgs/libcxx-4.0.1-h579ed51_0.tar.bz2
  expected md5 sum: 28471619a3a79da80d9b5747f1ec0221
  actual md5 sum: 291e61684d014641092020bdb1acb096
```

However, the changes that triggered the failed tests (https://github.com/PCMDI/cmor/commit/9f56aa6bf71c22e54315239c2e2f705d28cafa6a) didn't change the CircleCI configuration script.  The previous commit (https://github.com/PCMDI/cmor/commit/8ee711286de435522aaa5933ac848f1095389e91) that did change the CircleCI and Travis scripts was passing tests.

PR #500, opened 2019-06-12
Throw exception if not enough data is passed to cmor.write in Python

### LINKED ISSUE #485: not passing enough times as data passes
ntimes_written=2

but passing enough data only for 1 time should not work and right now it does

mauzey1:
Resolves #485

This change checks the amount of data being passed to cmor.write in Python if it is the expected amount for the amount of time passed (either by setting ntimes_passed or passing time values).  It will throw an exception if the amount is lower than expected (or if ntimes_passed is negative).

I also added a test to see if the exception is being thrown.

As for Fortran, it looks like it already handles this issue.
https://github.com/PCMDI/cmor/blob/8a48023d6456b596df5f0a1a9440bc3fe4d2634c/Src/cmor_fortran_interface.f90#L1032-L1037

### DIFF
--- a/Lib/pywrapper.py
+++ b/Lib/pywrapper.py
@@ -774,8 +774,15 @@ def write(var_id, data, ntimes_passed=None, file_suffix="",
     # Check if there is enough data for the number of times passed
     if ntimes_passed < 0:
         raise Exception("ntimes_passed must be a positive integer")
-    elif ntimes_passed > osh[0]:
-        raise Exception("ntimes_passed must be less than or equal to the number of times in data")
+
+    expected_size = ntimes_passed
+    for d in goodshape:
+        expected_size *= d
+    passed_size = 1
+    for d in sh:
+        passed_size *= d
+    if expected_size > passed_size:
+        raise Exception("not enough data is being passed for the number of times passed")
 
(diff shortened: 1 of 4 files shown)


mauzey1:
@doutriaux1 It is currently named test_python_CMIP6_CV_not_enough_data.py since the CMIP6_CV tests are the only Python tests currently being used in CI.  https://github.com/PCMDI/cmor/pull/493 will add the other Python tests back in.

I will get started on a Fortran test.

mauzey1:
@doutriaux1 Is there a way to catch an error in Fortran?  If cmor_write has an error it will return CMOR_CRITICAL, but I want to catch the "not enough data" message.

doutriaux1:
@mauzey1 I was dealing with these sort of things back here: https://github.com/PCMDI/cmor/pull/387/files

Let me know if it does not help you.

mauzey1:
@doutriaux1 I am not sure what to take from that.

Another issue with the Fortran test is that the error uses CMOR_CRITICAL, which exits the program at cmor_handle_error before cmor_write can return the error flag.

mauzey1:
@doutriaux1 I have moved the not-enough-data test from the CMIP6 CV tests to the Python tests in the Makefile.

PR #506, opened 2019-07-01
Check pressure level dimension size attribute for pressure levels test in PrePARE.

### LINKED ISSUE #502: wap_6hrPlev and zg_6hrPlevPt failing PrePARE
Hi all, 

sorry if I am opening a new issue on this but I am experiencing again the same problem I reported in #409 with PrePARE

Two variables `wap_6hrPlev` and `zg_6hrPlevPt` are still reporting the following errors when they are examined by PrePARE:

```
The entry wap could not be found in CMOR table
The entry zg could not be found in CMOR table
```

Looking at the corresponding cmor tables suggest that it is still the same kind of `out_name` issue (`wap` should be `wap4`, `zg` should `zg27` at least in our case): could be possible that these two are still not included in the correct list? 

thanks for any help!
Best,
Paolo

mauzey1:
Fixes #502

### DIFF
--- a/LibCV/PrePARE/PrePARE.py
+++ b/LibCV/PrePARE/PrePARE.py
@@ -394,6 +394,7 @@ class checkCMIP6(object):
                                           'filename': filename}):
                     # If test successfull, the CMOR entry to consider is given by the test
                     variable_cmor_entry = cmor_entry
+                    break
                 else:
                     # If not, CMOR entry to consider is the variable from filename or from input command-line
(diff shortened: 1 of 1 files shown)


mauzey1:
@doutriaux1 Python 3.6 doesn't have re.Pattern, which causes an error when running PrePARE with that version.  The last change makes it so any Python version below 3.7 will use re._pattern_type for PrePARE.

PR #526, opened 2019-07-30
Update Zenodo DOI

mauzey1:
@dnadeau4 The CircleCI tests are failing due to a conda server error but this change should not effect the build.  Merging it could get the nightly builds working since they also got hit with a conda server error.


dnadeau4:
I will try to rebuild first.

dnadeau4:
Things are going well.  I just approved this.  I tried your new DOI and it worked fine.
https://doi.org/10.5281/zenodo.3355583

PR #528, opened 2019-08-02
Clean up climatological coordinate bounds test

### LINKED ISSUE #328: generating/checking climatological coordinate bounds
The output requirements for CMIP6 ( https://goo.gl/neswPr ) provide guidance on how to treat the bounds on coordinates. It doesn't suggest that the bounds be treated differently if they represent a climatology (rather than a normal time axis) except the name of the bounds variable should be given by the CF "climatology" attribute rather than the CF "bounds" attribute.  The guidance states:

```
Usually, the bounds name should be the same as the coordinate name, but 
with the suffix “_bnds” appended.
```
The only CF requirement concerning climatological bounds is that the variable name should be recorded by the "climatology" attribute.

Given this, should PrePARE skip checking what the *name* of the climatological bounds variable is, and we can let the CF checker determine if the file meets CF requirements?

Concerning CMOR3, it would be more consistent with the output requirements document if the name of climatological bounds were the same as the name of the regular bounds, so "time_bnds".  Should we make that change?

mauzey1:
Resolves #328


taylor13:
I think I'm supposed to review this.  I looked at Resolves #328 , and I see no reason not to approve the pull request.

PR #531, opened 2019-08-09
Make git ignore build-generated and test-generated files

mauzey1:
These changes will make git ignore build files and tests executables.  They also change some tests that were creating their own tables and log files.  The tests will now delete their generated tables when they are done with them, and will no longer generate log files.


taylor13:
Was there no useful information in the log files?  Just wondering.

mauzey1:
@taylor13 The log file was storing the execution times for cmor and cdms2 for different deflate and shuffle values.  These values were already being printed to the console.  I could make the tests output a log file again but with a flag that prevents it from being created by default.

taylor13:
o.k. thanks for explaining.  Thanks for cleaning things up.

mauzey1:
@doutriaux Could you review this PR?  Thank you.

mauzey1:
@durack1 I could also add @sterlingbaldwin but his name doesn't appear in the list of reviewers for CMOR.

durack1:
@mauzey1 let's fix that, @sterlingbaldwin was just added. You're a repo admin so you should be able to manage the repo users - I also wonder whether we still need access for @dnadeau4 and wachslyon?

mauzey1:
@durack1 @dnadeau4 has left PCMDI but @wachsylon occasionally contributes to this project.

PR #564, opened 2019-11-23
Pin openblas 0.3.6 and numpy >=1.14.6 in meta.yaml

mauzey1:
The nightly builds for OSX were trying to use an older version of numpy that was not compatible with the other packages.  This change will pin numpy versions 1.14.6 and greater to the nightly build.

It will also pin openblas 0.3.6 for compatibility with cdms2 from conda-forge.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -77,9 +77,9 @@ aliases:
   - &conda_upload
     name: conda_upload
     command: |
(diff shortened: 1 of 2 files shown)


mauzey1:
I have made a new PR that will use the nightly build of CDMS2 and pin libnetdf 4.6.2 instead.

PR #566, opened 2019-12-13
Pin libnetcdf 4.6.2 and numpy >= 1.14.6

mauzey1:
Closes #564 

The OSX builds in CircleCI have been downloading numpy 1.11, which is outdated and causing errors in the build.  The PR will pin numpy version 1.14.6 and greater.

This PR will also pin libnetcdf 4.6.2 so that we can use the nightly CDMS2 builds again.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,7 @@ aliases:
        conda config --set always_yes yes --set changeps1 no
        conda update -y -q conda
        conda config --set anaconda_upload no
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -57,7 +57,7 @@ aliases:
        export UVCDAT_ANONYMOUS_LOG=False
        # run tests again but with cdms2 installed
        set +e
-       conda install -q -n py$PYTHON_VERSION -c conda-forge cdms2
+       conda install -q -n py$PYTHON_VERSION -c cdat/label/nightly -c conda-forge cdms2
```
durack1: @mauzey1 I assume there's no problem in using a `cdms nightly` build for these tests?

mauzey1: @durack1 We were originally using the nightly build of cdms2 for cmor, but I decided to switch to the conda-forge version since it was using a later version of libnetcdf.  However, I decided to switch back to the nightly version of cdms2 and pin libnetcdf 4.6.2 since that version is pinned for cdms.

PR #567, opened 2019-12-19
Updates for nwdFracLut test

### LINKED ISSUE #535: nwdFracLut in Emon
Hi,
when I try to produce `nwdFracLut` from `Emon`, I receive
```
C Traceback:
! In function: 

!!!!!!!!!!!!!!!!!!!!!!!!!
!
! Error: You are defining variable 'nwdFracLut' (table Emon)  with 4 dimensions, when it should have 5
!
!!!!!!!!!!!!!!!!!!!!!!!!!

```
after I call `cmor_variable`.
I skip the registration of axis `typenwd` since it is a singleton.

Are you able to produce nwdFracLut or reproduce the error?

Best regards,
Fabi

mauzey1:
Closes #539 
Resolves #535 

This PR fixes the issues that I discussed in https://github.com/PCMDI/cmor/issues/535#issuecomment-544771072 that were causing the nwdFracLut test to fail.


mauzey1:
@taylor13 The NetCDF dimension "strlen" will get defined with the greatest string length of all string axes and singletons if a string axis is found in the function `cmor_define_dimensions`.  If it is not defined in that function, then "strlen" will be defined with the greatest string length of all string singletons in the function `create_singleton_dimensions`.

taylor13:
sounds good to me.

Issue #495, opened 2019-06-03
Error that doesn't cause failure when producing mrsos for Lmon


ehogan:
I'm using CMOR 3.4.0 to produce `mrsos` for `Lmon`. The output netCDF file is produced successfully, yet the CMOR log contains:
```
C Traceback:
! In function: cmor_treat_axis_values
! called from: cmor_axis
! 

!!!!!!!!!!!!!!!!!!!!!!!!!
!
! Error: axis sdepth1 (table: Lmon) , detected value at:
! 0.100000 when valid_max is 0.100000
!
!!!!!!!!!!!!!!!!!!!!!!!!!
```
The bounds I'm providing for `sdepth1` are `[0, 0.100000001490116]` (this is what is written to the output netCDF file). Should [cmor_treat_axis_values](https://github.com/PCMDI/cmor/blob/master/Src/cmor_axes.c#L1240) accept that `0.100000001490116` is close enough to `0.1` by performing an appropriate float comparison instead of producing this error (that doesn't cause CMOR to fail anyway)?

taylor13:
I think we should either allow the value and not generate an error message, or error exit and genera te the message.  It doesn't seem to me that this would be too difficult in theory.  When it finds the error, why don't we throw an error?  (i.e., I don't think we should allow it)   Does the problem have to do with a conversion from single to double precision?

ehogan:
> I think we should either allow the value and not generate an error message, or error exit and generate the message. 

Agreed :)

> Does the problem have to do with a conversion from single to double precision?

It's a float representation issue:
```
% python
Python 2.7.15 |Anaconda, Inc.| (default, Dec 14 2018, 19:04:19) 
[GCC 7.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy as np
>>> my_float = np.float32(0.1)
>>> print('{0:.15f}'.format(my_float))
0.100000001490116
```
The line I linked to in my first comment performs a `if values > valid_max` test; this line (and others like it) should be adapted to ensure this is taken into account :)

taylor13:
It seems to me that valid_max and valid_min when applied to (or compared with) an "axis" value should be declared as "double precision", but it appears they have been declared single precision (with a value of +- 1.e20).  I don't know c syntax, but in fortran that's a single precision float.  If I'm right, this should be the first thing we correct.

[...]

PR #577, opened 2020-02-01
Allow axis values to be within 1.e-6 fractional error of the valid_min and valid_max

mauzey1:
#495 
Allow axis values within a fractional error of the valid_min or valid_max of an axis.  This is designed to allow values that are meant to be the min or max of an axis but have been converted from floats to doubles.


durack1:
@mauzey1 I am about to take off on international travel and then some vacation, for PRs over the next couple of weeks, I suggest pinging @taylor13 alone for a code review

taylor13:
As we discussed, we should change the "<" to "<=" and ">" to ">=".  Otherwise, when the min or max value is 0 and the user-passed value is 0, the code will raise an error when the value of 0 is perfectly acceptable.

mauzey1:
@taylor13 These lines are meant to detect when the value goes beyond the threshold, so they shouldn't raise an error when the value and the min/max are zero.
https://github.com/PCMDI/cmor/blob/a246d1d0eb6cf5f5f42920d951c0564e06233b8a/Src/cmor_axes.c#L1196-L1211
https://github.com/PCMDI/cmor/blob/a246d1d0eb6cf5f5f42920d951c0564e06233b8a/Src/cmor_axes.c#L1222-L1225
https://github.com/PCMDI/cmor/blob/a246d1d0eb6cf5f5f42920d951c0564e06233b8a/Src/cmor_axes.c#L1240-L1255
https://github.com/PCMDI/cmor/blob/a246d1d0eb6cf5f5f42920d951c0564e06233b8a/Src/cmor_axes.c#L1262-L1265

taylor13:
My mistake ... don't know why I didn't correctly interpret the logic in the case of a min/max = 0.  Obviously there is need for any action.

PR #581, opened 2020-02-21
Improve error messages for supported data types in pywrapper

### LINKED ISSUE #542: Error message mis-match
The error message [here](https://github.com/PCMDI/cmor/blob/2e9f6b2f6acbd25d15b6c46a26176632c864ac8b/Lib/pywrapper.py#L429) should match the if condition on the previous line.

mauzey1:
Resolves #542

### DIFF
--- a/Lib/pywrapper.py
+++ b/Lib/pywrapper.py
@@ -47,16 +47,16 @@ def set_terminate_signal(signal):
 
 
 def time_varying_grid_coordinate(
-        grid_id, table_entry, units, type='f', missing_value=None):
+        grid_id, table_entry, units, data_type='f', missing_value=None):
     """ Create a cmor variable for grid coordinates in case of time varying grids
     Usage:
     coord_grid_id = grid_time_varying_coordinate(
-    grid_id, table_entry,units,type='f',missing_value=None)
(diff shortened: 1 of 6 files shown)


Lib/pywrapper.py
```diff
@@ -154,7 +154,7 @@ def grid(axis_ids, latitude=None, longitude=None,
         nvert = 0
         if not type in ['d', 'f', 'i', 'l']:
```
sterlingbaldwin: the word "type" is a built in python method and should not be used for variable names. Although its technically not a reserved word, this is a bad style choice. I know the people that originally wrote this code arent here any more, but that doesnt mean it cant be fixed!

mauzey1: @sterlingbaldwin @durack1 
I have changed the 'type' variable in pywrapper.py to 'data_type' to not shadow the built-in 'type' in Python.  This also involved changing the parameters of some functions to reflect this change.  I will add these changes to the CMOR documentation site.

sterlingbaldwin: Thanks, I know it may seem small, but this kind of thing can cause serious headaches.

PR #583, opened 2020-03-04
Fix count of files with errors in PrePARE

### LINKED ISSUE #540: Recursive directory output issue / error vs warning
I've ran PrePARE to recurse a subdirectory.  When this happens, some messages are ambiguous regarding which file they pertain to.  These for the most part must be `warning` messages but its isn't entirely clear.  One such message is the following, and I ran again after listing the files and providing each via a sh loop:

ESC[2;31;47mC Traceback:
! In function: _CV_checkParentExpID
! ESC[0m

ESC[1;31;47m!!!!!!!!!!!!!!!!!!!!!!!!!
!
! Error: Your input attribute "parent_mip_era" is not defined 
! properly for historical 
! Please refer to the CMIP6 documentations.
! 
!
!!!!!!!!!!!!!!!!!!!!!!!!!ESC[0m

ESC[95m
Number of files scanned: 1ESC[0mESC[1;32m
Number of file with error(s): 0ESC[0m

The message indicates an error, but the summary info at the bottom says otherwise.  Is the message really a warning, or should the bottom count it as an Error?

mauzey1:
Resolves #540

### DIFF
--- a/LibCV/PrePARE/PrePARE.py
+++ b/LibCV/PrePARE/PrePARE.py
@@ -447,14 +447,14 @@ class checkCMIP6(object):
         # -------------------------------------------------------------------
         # Check global attributes
         # -------------------------------------------------------------------
-        self.errors += cmip6_cv.check_requiredattributes(table)
-        self.errors += cmip6_cv.check_institution(table)
-        self.errors += cmip6_cv.check_sourceID(table)
-        self.errors += cmip6_cv.check_experiment(table)
-        self.errors += cmip6_cv.check_grids(table)
-        self.errors += cmip6_cv.check_ISOTime()
-        self.errors += cmip6_cv.check_furtherinfourl(table)
-        self.errors += cmip6_cv.check_subExpID(table)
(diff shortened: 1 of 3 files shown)


mauzey1:
These changes are meant to resolve the problem that @sashakames encountered where PrePARE didn't increment its count of files with errors when processing a file with the missing attribute `parent_mip_era`.

These changes include:
* Making errors detected in `cmor_CV_checkParentExpID` return a value of -1.  There were several error checks that didn't have a return value.
* Increasing the error count in PrePARE by checking if the error code returned by the cmip6_cv functions are non-zero rather than adding the error code to the count.  This prevents different error codes from clobbering each other causing erroneous results.
* Modify the test [test_python_CMIP6_CV_parenttimeunits.py](https://github.com/PCMDI/cmor/blob/92c04c3d6c5288c1c631c2d2e5b58472427e260f/Test/test_python_CMIP6_CV_parenttimeunits.py) to handle the exception thrown when `cmor.write` stops due to an error detected in `cmor_CV_checkParentExpID`.

PR #586, opened 2020-03-12
Check if the path 'include/udunits2' exists for udunits2 include path

### LINKED ISSUE #545: needs to test for subdirectory usage in udunits include path
The `udunits2-devel` RPM from EPEL uses paths `/usr/include/udunits2/*.h` rather than `/usr/include/*.h`.

During build, this leads to:
```
gcc -c -g -DCOLOREDOUTPUT -I./include -I./include/cdTime  -lnetcdf -g -O2 -lnetcdf -ludunits2    Src/cmor.c
In file included from ./include/cmor.h:578:0,
                 from Src/cmor.c:9:
./include/cmor_func_def.h:8:22: fatal error: udunits2.h: No such file or directory
```

A successful build is possible using
```
make UDUNITS2FLAGS='-I/usr/include/udunits2 -ludunits2'
```

Could this issue be tested for in the configure step, and the include path set automatically, so that this workaround is not necessary?

mauzey1:
Resolves #545


durack1:
@mauzey1 just spotted this is unreviewed - do you need this looked at, or should it be closed?

@sashakames @taylor13 ping

mauzey1:
@durack1 @alaniwi
This pull request has been sitting here for a long time.  It was meant to address an issue where the header file for udunits2 needed to be found in `/usr/include/udunits2/` instead of `/usr/include/`.  This issue relevant to a RPM installation udunits2, which installs the header file differently from our Conda build environment method.

I haven't heard anything from @alaniwi since then about this.  I think I will just accept this change since it doesn't seem to break anything else in the build.

sashakames (approved):

PR #596, opened 2020-04-22
Python 3.8 support

mauzey1:
Adding support for Python 3.8 to CMOR.

Some changes needed to be made in _cmormodule.c for changes in 3.8.  `PyCMOR_zfactor` was having issues with list of axis ids that had a length of one as it was trying to read them as scalars using `PyLong_AsLong`.  This was causing an error in this section of a test.
https://github.com/PCMDI/cmor/blob/2b8af958b0615c975005dfec361edd3f3ce67c35/Test/test_python_user_interface_00.py#L116-L120
To fix this, I replaced checking if the number of axis ids was greater than 1 with checking if the axis id object was a scalar using `PyArray_CheckAnyScalar`.
https://github.com/PCMDI/cmor/blob/53d53fc51246556b3a406c12bfd2091792fc9078/Src/_cmormodule.c#L674-L686
There was a similar section in `PyCMOR_write` with the number of times passed.  This was not causing issues with tests but I applied the same logic as with the zfactor axis ids to the time passed.  This is just in case `PyFloat_AsDouble` exhibits the same behavior.
https://github.com/PCMDI/cmor/blob/53d53fc51246556b3a406c12bfd2091792fc9078/Src/_cmormodule.c#L827-L834
From https://docs.python.org/3/whatsnew/3.8.html

> Functions that convert Python number to C integer like PyLong_AsLong() and argument parsing functions like PyArg_ParseTuple() with integer converting format units like 'i' will now use the __index__() special method instead of __int__(), if available. The deprecation warning will be emitted for objects with the __int__() method but without the __index__() method (like Decimal and Fraction). PyNumber_Check() will now return 1 for objects implementing __index__(). PyNumber_Long(), PyNumber_Float() and PyFloat_AsDouble() also now use the __index__() method if available. (Contributed by Serhiy Storchaka in bpo-36048 and bpo-20092.)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -166,6 +166,27 @@ jobs:
       - run: *run_prepare_tests
       - run: *conda_upload
 
+  macos_cmor_py38:
+    macos:
+      xcode: "11.1.0"
+    environment:
+      PYTHON_VERSION: "3.8"
+      WORKDIR: "workspace/test_macos_cmor"
+      CONDA_COMPILERS: "clang_osx-64>=9 gfortran_osx-64>=7"
+      LDSHARED_FLAGS: "-bundle -undefined dynamic_lookup"
+      UVCDAT_ANONYMOUS_LOG: "False"
+      OS: "osx-64"
+    steps:
+      - checkout
+      - run: *pull_submodules
+      - run: *setup_miniconda
+      - run: *create_conda_env
+      - run: *setup_cmor
+      - run: *run_cmor_tests
+      - run: *run_cmor_tests_with_cdms2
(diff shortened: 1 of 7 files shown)


markcollier:
Great @mauzey1 I've been waiting for this day! Can likely put python2 to pasture now...I have applied this version to a few cases and without issues apart from address valid cmor error messages. Thanks

durack1:
@markcollier welcome to the future!

Just an FYI, py3 support was rolled out on the 30 July 2019, see https://anaconda.org/conda-forge/cmor/files?page=1

PR #599, opened 2020-04-30
Update version number to 3.6.0

mauzey1:
Update version number of CMOR to 3.6.0.


configure.ac
```diff
@@ -2,7 +2,7 @@ dnl
 dnl Process this file with autoconf to produce a configure script.
 
 dnl AC_PREREQ(2.59)
-AC_INIT(cmor, 3.5.0, doutriaux1@llnl.gov)
+AC_INIT(cmor, 3.6.0, doutriaux1@llnl.gov)
```
durack1, line 5: @mauzey1 you should update the contact email here

configure
```diff
@@ -580,8 +580,8 @@ MAKEFLAGS=
 # Identity of this package.
 PACKAGE_NAME='cmor'
 PACKAGE_TARNAME='cmor'
-PACKAGE_VERSION='3.5.0'
-PACKAGE_STRING='cmor 3.5.0'
+PACKAGE_VERSION='3.6.0'
+PACKAGE_STRING='cmor 3.6.0'
 PACKAGE_BUGREPORT='doutriaux1@llnl.gov'
```
durack1, line 585: Same here

durack1:
@mauzey1 we have to be a little careful with the information updates at https://github.com/PCMDI/cmip6-cmor-tables/compare/fa22c10345b0cbb7d002bb81ac1865c775340be8...4422bd5c83bb46bf3301716edb31798baa5fd2ae#diff-c43e4fa1a1a7069a07deaf63c70f97a7R4

So the `"cmor_version": "3.5",` -> `"cmor_version": "3.6",` as I am not sure this will allow older table files cmor_version < 3.6 to work. Has this been checked?

@taylor13 ping

configure
mauzey1: I have changed the email from @doutriaux1 to mine.

configure
```diff
@@ -580,9 +580,9 @@ MAKEFLAGS=
 # Identity of this package.
 PACKAGE_NAME='cmor'
 PACKAGE_TARNAME='cmor'
-PACKAGE_VERSION='3.5.0'
-PACKAGE_STRING='cmor 3.5.0'
-PACKAGE_BUGREPORT='doutriaux1@llnl.gov'
```
doutriaux1, line 585: @mauzey1 thanks!

mauzey1:
@durack1 I have reverted the change to the `cmor_version` attribute of the CMIP6 tables.  I was under the assumption that they needed to be updated when CMOR was updated.

durack1:
@mauzey1 thanks for this, I vividly recall this causing issues with previous changes from 3.4 -> 3.5, I don't recall if the logic within CMOR was change as a consequence. I'll wait for @taylor13 to chime in, as I was only peripherally involved with that change

durack1:
@taylor13 is there an issue in incrementing the `cmor_version`?

taylor13:
I agree that we should check what the logic is in CMOR. Changes to the CMOR version from 3.5 to 3.6 does seem appropriate because significant changes are made to the code.  And we haven't made changes to CMOR that might prevent it from reading older .json files.  On the other hand, if CMOR is at some point in the future generalized so that it can *also* read input files formatted differently, then *those* new-form input files should specify that they cannot be read unless you use the updated version of CMOR.  

My understanding of the intent of the "cmor_version" specification in each of the .json variable files is that it indicates the earliest version of cmor that it could be interpreted by.  So if the .json file indicates "3.5" for the cmor_version, it should be readable by version 3.5, 3.6, and any successive versions (perhaps up to cmor 4.0?? when we might allow cmor to become non-compatible with earlier .json files).

The logic in the cmor check should not require the cmor_version number to be *identical* to the actual version of cmor it is being read by.

I suspect CMOR behaves as I think it should, but given Paul's memory, we should check.

As usual, there be something I'm missing here.

mauzey1:
@taylor13 @durack1 
The logic in CMOR for checking its version against the table's cmor_version is posted below.  It throws an error if the table version is greater than CMOR's version.
https://github.com/PCMDI/cmor/blob/785b3904844be2fbadddf2c8a602788d4ac7d045/Src/cmor_tables.c#L332-L351

durack1:
@mauzey1 thanks for checking. So this means if we increment the [cmip6-cmor-tables](https://github.com/PCMDI/cmip6-cmor-tables/) `cmor_version` entries, we break all currently available CMOR <3.6 versions using the latest tables

taylor13:
@mauzey1 In checking this, I'm struggling with not knowing what the variables represent in the following code segment:
---
    if (strcmp(att, TABLE_HEADER_VERSION) == 0) {
        d2 = CMOR_VERSION_MAJOR;
        d = CMOR_VERSION_MINOR;
        while (d > 1.)
            d /= 10.;
        d2 += d;
        sscanf(value, "%f", &d);
        if (d > d2) {
---
I assume `TABLE_HEADER_VERSION` contains the string, "cmor_version" and for the latest .json files, so we now want to consider whether the version recorded in the table (currently "3.5") is compatible with the CMOR code version that is reading it, which presumably is recorded by, CMOR_VERSION_MAJOR = 3 and CMOR_VERSION_MINOR = 6 (in this latest release).  If this is correct, then isn't the result of the "while" loop segment d=.6 and d2=3.6?   
Next I guess d=.5 is replaced with the version number obtained from the .json file:  d=3.5
And then if d > d2.  (In our case d=3.5 and d2=3.6, so it won't error exit.  
If I've read the code correctly, then I'd say it behaves exactly as I had hoped.  It will only error exit if the version in the table exceeds the CMOR code version trying to read it.  Please check my logic and if you need me to approve something, let me know.

mauzey1:
@taylor13 Yes, that is exactly what that code segment is doing.

Aside from this pull request to update the version number, there is #600 that has you as a reviewer.

PR #600, opened 2020-05-02
Fixing issues in cmor_CV.c

### LINKED ISSUE #245: Error handling in cmor_CV.c
Some of the functions in the [cmor_CV.c file](https://github.com/PCMDI/cmor/blob/master/Src/cmor_CV.c) return a value of `-1` when an error is encountered, while others return a value of `1`.

This causes problems [at this point in the cmor.c file](https://github.com/PCMDI/cmor/blob/master/Src/cmor.c#L2854), since the return values from each call to the `cmor_CV_check*` functions are added together. Depending on the combination of issues with the CVs in the `cmor_dataset` JSON file, this can result in the returned `ierr` being `0`.

Would it be possible to use a consistent return value (i.e. `1` so the `ierr` used in the [error message](https://github.com/PCMDI/cmor/blob/master/Src/cmor.c#L4305) indicates the number of errors) when an error is encountered for the functions in [cmor_CV.c file](https://github.com/PCMDI/cmor/blob/master/Src/cmor_CV.c), please?

mauzey1:
Closes #245 

This PR applies fixes to issues identified in #245 

* `cmor_CV_checkSourceID`: Throw an error if `source` is not found in a `source_id` section of the CV.
* `cmor_CV_checkParentExpID`: If `parent_experiment_id` is "" or not found in the variable file, then it will be set to "no parent" if that is the valid value for `parent_experiment_id` in the CV.
* `cmor_CV_checkParentExpID`: When `parent_experiment_id` is a value other than "no parent", then process `parent_activity_id`, `branch_method`, `branch_time_in_child`, `branch_time_in_parent`, `parent_time_units`, `parent_variant_label`, `parent_source_id`, and `parent_mip_era` and only return an error after going through all of them.


durack1:
@mauzey1 as I have not been involved in these discussions, it makes little sense for me to review this. I will leave this to @taylor13

taylor13:
will get to this later today or (more likely) tomorrow.  Sorry for the delay.

taylor13:
I'm afraid that my python skills are too weak to review and understand what the code is now doing.  Above you say:

1. cmor_CV_checkParentExpID: If parent_experiment_id is "" or not found in the variable file, then it will be set to "no parent" if that is the valid value for parent_experiment_id in the CV.

2. cmor_CV_checkParentExpID: When parent_experiment_id is a value other than "no parent", then process parent_activity_id, branch_method, branch_time_in_child, branch_time_in_parent, parent_time_units, parent_variant_label, parent_source_id, and parent_mip_era and only return an error after going through all of them.

I think the first point above is the proper behavior if "found in the variable file" means "found in the file containing all user directives concerning the dataset to be written (e.g., CMIP6_input_example.json)".  [We should agree on a name for this like "user-defined CMOR control file".]    

In addition to the check executed in point 1: a) if "no parent" is *not* a valid option for parent_experiment_id, a fatal error should be raised; b) all of the other "parent_experiment" information strings ( `parent_activity_id`, `parent_source_id`, `parent_variant_label`, `parent_time_units`, `branch_method`) should be omitted or, if you prefer, set to "no parent" (and unless the user has specified "" or "no parent" for these attributes or simply omitted any specification of them, a warning should be raised); similarly the `branch_time_in_child`, and `branch_time_in_parent` should be omitted.

With regard to the second point above, this looks like what we want if by "process" you mean "check that the attributes have been defined and check that `parent_activity_id` is consistent with `parent_experiment_id`".

It's possible that the code isn't quite as thorough as this, but it should be consistent with this.  If you think it is, then we should approve the pull request.

Thanks.

mauzey1:
@taylor13 

By "found in the variable file", I mean "found in the NetCDF file processed by PrePARE".  If the NetCDF file being processed by PrePARE has the attribute `parent_experiment_id` set to "", or the attribute is missing from the file, and CMIP6_CV.json has "no parent" for its `parent_experiment_id`, then the NetCDF file will be treated as if it had `parent_experiment_id = "no parent"`.  Otherwise, if  `parent_experiment_id` is _not_ "no parent" in CMIP6_CV.json, then an error will be thrown.

As for checking that `parent_activity_id` is consistent with `parent_experiment_id`, I have notice these situations in CMIP6_CV.json listed below.

https://github.com/PCMDI/cmip6-cmor-tables/blob/2814801a704afcf47e4ad3888f0b0cd994d3ffab/Tables/CMIP6_CV.json#L3895-L3913 
https://github.com/PCMDI/cmip6-cmor-tables/blob/2814801a704afcf47e4ad3888f0b0cd994d3ffab/Tables/CMIP6_CV.json#L4704-L4721

For experiment "dcppA-hindcast-niff", would it be okay to have `parent_experiment_id` be "dcppA-assim" and `parent_activity_id` be "no parent"?

taylor13:
thanks for correcting my misunderstanding.  I thought this code was executed before CMOR wrote a variable to a file.  I didn't know this was a "PrePARE" check.  That makes a big difference.  I'll re-review in that light and get back to you.  PrePARE can't be as strict (or insistent) as CMOR (pre-write stage).

mauzey1:
@taylor13 This code is used by both cmor_write and PrePARE.  I use PrePARE checking as the more convenient example, but cmor_write should encounter the same issues with user-defined input.

taylor13:
I see.  So does CMOR execute it both before it writes the output and then again when it's running through the PrePARE checks, or if CMOR has just written the file, does it skip redoing some (or all?) of the PrePARE checks on the written file?

I was thinking that when CMOR does the checks before writing the files, then it could auto-correct some of the attributes that weren't quite right (issuing a warning that it is doing this).  But if we want the exact same code to run on files already written (not through CMOR), then we wouldn't be able to include any auto-corrections (because we don't want PrePARE to actually alter any written files.

mauzey1:
The CV checks are performed during the first part of cmor_write, before a file gets written.  When it is being used by PrePARE, it is only being used to detect errors in the file not overwrite them.

mauzey1:
@taylor13 Do you have any more questions that need answering before approving this PR?

taylor13:
I'm reviewing this issue now.  I just noticed, I didn't respond to:
```
For experiment "dcppA-hindcast-niff", would it be okay to have parent_experiment_id 
be "dcppA-assim" and parent_activity_id be "no parent"?
```
The answer is no.  The parent_activity_id must be one of the parent_activity_id's listed for that parent_experiment_id, but it also must be consistent with the parent_experiment itself.  (The activities associated with a parent_experiment_id are the same as the activities associated with the experiment_id's recorded in [CMIP6_experiment_id.json](https://github.com/WCRP-CMIP/CMIP6_CVs/blob/master/CMIP6_experiment_id.json)).  You'll note that none of the activity_id's listed there are "no parent", but some of the parent_experiment_id's list "no parent" as an option (sometimes the only option).  That's because *every* experiment has at least one activity sponsoring it, but some experiments don't have a parent.

mauzey1:
@taylor13 The current changes don't catch the situation discussed in https://github.com/PCMDI/cmor/pull/600#issuecomment-629367183.  Should I add another change to handle that situation, or just merge the current changes?

taylor13:
I would just go ahead and merge.  We might transfer that to a new issue to be considered in a future release.  Thanks.

PR #616, opened 2020-11-05
Properly handle unicode values passed to cmor.axis in Python

### LINKED ISSUE #612: cmor.axis is not handling unicode string values correctly in Python
This is a followup of #611.

Passing axis values of unicode string type to `cmor.axis` will cause CMOR to store only the first characters of the axis values.

Example:
```
#Values passed to CMOR
sector_values = [
                  'barents_opening', 
                  'bering_strait',
                  'windward_passage', 
                  'davis_strait', 
                  'denmark_strait', 
                  'drake_passage',
                  'english_channel', 
                  'faroe_scotland_channel', 
                  'florida_bahamas_strait',
                  'fram_strait', 
                  'gibraltar_strait', 
                  'iceland_faroe_channel', 
                  'indonesian_throughflow', 
                  'mozambique_channel', 
                  'pacific_equatorial_undercurrent', 
                  'taiwan_luzon_straits'
                ]

sector_string = np.array(sector_values, dtype='U')
oline_axis_id = cmor.axis(table_entry='oline', units='kg s-1', length=len(sector_string), coord_vals=sector_string)

...
#What gets stored in the NetCDF file
 sector =
  "b",
  "b",
  "w",
  "d",
  "d",
  "d",
  "e",
  "f",
  "f",
  "f",
  "g",
  "i",
  "i",
  "m",
  "p",
  "t" ;
```
This can be fixed by converting the unicode strings to zero-terminated byte strings, which is type `'S'` in numpy.
```
sector_string = np.array(sector_values, dtype='S')
```
Shouldn't this conversion happen in CMOR?  Or should we change the code and documentation to say that only strings of type `'S'` are accepted by `cmor.axis`?

mauzey1:
Fixes #612 

When string values get passed to cmor.axis, they will be treated as a Numpy array of type numpy.unicode_.

From https://numpy.org/doc/stable/reference/arrays.dtypes.html
> Note that str refers to either null terminated bytes or unicode strings depending on the Python version. In code targeting both Python 2 and 3 np.unicode_ should be used as a dtype for strings.

The C code has been modified to properly copy string values from a Numpy array to a C array to be processed by cmor_axis.


durack1:
@mauzey1 I wonder if explicitly decoding to unicode e.g.
```python
string.decode('utf-8')
```
Would be a good way to force all types?

mauzey1:
@durack1 
The current line below should already be performing this conversion.
```
coord_vals = numpy.array(coord_vals, numpy.unicode_)
```

I could make the conversion more explicit by using
``` 
if data_type == 'S':
    coord_vals = numpy.char.decode(coord_vals, encoding='utf-8')
```

durack1:
@mauzey1 my comment was just that. I found that in generating html content from the contributed contents for the [CMIP6_CVs](https://github.com/WCRP-CMIP/CMIP6_CVs) all manner of weird non-standard (and non UTF-8) characters were sneaking in, and the parsers weren't expecting these characters and consequently barfed.

I think this is true to CMOR, so catching non UTF-8 characters and throwing an explicit error (if a decode function can't handle things) would be my more bulletproof preference

durack1 (approved):
@mauzey1 happy to green light your contribution, it looks good, but was just wanting to make the statement about non-UTF-8 chars for the record

mauzey1:
@durack1 After some experimenting with unicode, I decided to change the code to explicitly convert the strings to UTF-8.

durack1:
@mauzey1 I think that was a wise tweak, hopefully, it catches those fringe cases

Issue #623, opened 2021-04-27
enhancement for forecast data


taylor13:
The decadal prediction community have proposed an extension of CMIP6 standards to accommodate additional metadata when forecasts are stored by CMOR.  Their proposal can be found at https://docs.google.com/document/d/1T-Xlkc07kzDbtyFp-JJiRQYmHO-OzFOgXjCvFGW5M7Y/edit .  

From a brief look at that document, it appears that it will be a non-trivial enhancement that might include:

1. Currently CMOR attaches a "coordinates" attribute to a variable for the following cases:
- scalar coordinates (e.g. variable tas has a scalar coordinate "height" where the height of the surface air temperature is stored)
- coordinates when text string labels are needed (e.g., for variables that are a function of some index, like the index used to distinguish between different ocean basins.  In this case the ocean basin name is stored in an auxiliary coordinate associated with the variable through the "coordinates" attribute.
- latitude and longitude values for grids other than cartesian latitude x longitude grids.

(See https://goo.gl/neswPr for further description of the CMIP6 requirements.)
For forecast data the "coordinates" attribute should include "reftime" and "leadtime" in its list.  The "reftime" is a scalar coordinate and can probably be treated by CMOR just like the other scalar coordinates except it probably should not have a default value specified in the CMOR table.  Also, CMOR may be currently limited to handling a single scalar coordinate for a given variable (not sure about this), so this might be an extension. 

Storing the "leadtime" will likely require more work.  This auxiliary coordinate variable must be the same size (length) as "time" and contains the time elapsed since the start of the forecast measured in days.  It is equivalent to "time" except for the reference time is specified in the time coordinate's units attribute, whereas the reference time for "leadtime" is stored in "reftime".  These two reference times will generally be different, so the "leadtime" will be offset from "time" by a constant amount.  CMOR could generate the "leadtime" values, given the "time" coordinate values, by simply subtracting from "time" the difference between "reftime" and the reference time stored in the units attribute of "time".  (Given this close and known relationship between leadtime and time, I wonder why "leadtime" is needed; wouldn't "reftime" be sufficient?)  

2. The format for "sub_experiment" has been extended to include the "day" as well as year and month when the experiment was initiated.  This seems like a minor change that should be easy to accommodate.  

3. There are additional global attributes called for that include free text supplied by the user.  I think CMOR can already handle these.

I'll try to get Paco to look this over and see if I've covered the requirements.  Also, perhaps he can explain the rationale for including "leadtime".

[...]

PR #634, opened 2021-10-06
implement lead time coordinate

piotr-florek-mohc:
This is my first go at issue #623. Adding scalar reference time could be achieved just by specifying a value of `reftime1` in `mip_tables`, which can be later overridden during runtime (please note that `reftime` is a restricted keyword in the cfunits namespace).

Adding leadtime has been achieved by implementing a simple function which creates an auxiliary lead time coordinate, and populates it with differences between time coordinate values and the reftime. Calendars and units should be the same, and the reftime should be outside the bounds of the time coordinate (i.e. must be earlier than the opening bound of the first time step).

I've also added a small tweak to `cmor_write()` function, to allow me to retrieve the `ncid` after the write is complete.

cc @matthew-mizielinski


piotr-florek-mohc:
Hi Karl,
apologies, I was meant to answer other questions yesterday but got distracted... 

> "reftime1" in the coordinates table should have "must_have_bounds = "no" and bounds_values should be set to the null string. I think it's also probably a mistake to define a default value for "reftime1", since this will probably vary from one forecast to the next.

This is just a placeholder value, as it's what triggers coordinate's interpretation as a scalar in CMOR's code. I think it would require much more work to change this behaviour, and make it backwards compatible at the same time, although I could probably add both reftime and leadtime in one go, as a separate API.

I have removed bounds from the CV_coordinates file (sorry, missed that one yesterday).

> [Also, although reftime may be a restricted key word in cftime, that shouldn't preclude its use as the identifier for the reftime coordinate variable in CMOR should it? If it shouldn't, then what is the reason for naming it "reftime1" instead of simply "reftime"?

The cftime issue was a bit of a conjecture on my part (I investigated it early on, when I wasn't familiar with CMOR's code at all, and I vaguely remember the namespace conflict was present in Python netCDF4). The fact is that CMOR segfaults during writing of the variable to the file (in `cmor_write_var_to_file()`) when 'reftime' is used, but does not when a different name is provided in mip tables instead. I've tried to debug this behaviour but I still have no clue why this happens.

> Normally a forecast field will be a synoptically sampled ("instantaneous") meteorological field, in which case "must_have_bounds" should be set to "no" for leadtime. If some of the forecast fields should be accumulated or "time-means" and others will be instantaneous snapshots, then 2 different "leadtime" coordinates should be defined (e.g. analogous to "time" and "time1").

Ahh I see. Okay, I will need to have a think how to resolve this.

> Does the presence of "reftime1" in the list of "coordinates" appearing with a variable in the amon.json file trigger CMOR to create leadtime and store both leadtime and reftime in the file (and include both of these variables in the netcdf "coordinates" attribute list?

Only if the new function (`calculate_leadtime_coord`) is explicitly called after reftime was created.

mauzey1:
@piotr-florek-mohc You will want to merge the latest from master to remove the Python 3.6 tests since we are no longer supporting Python 3.6 in an upcoming release.

piotr-florek-mohc:
thanks @mauzey1, that has been done

mauzey1:
@piotr-florek-mohc I noticed that you included a Python test but it is not part of the continuous integration test suite.  Please add the following line to the file Makefile.in under the `test_python: python` section:
```
test_python: python
	env TEST_NAME=Test/test_python_forecast_coordinates.py make test_a_python
```
I tried running the test but it experienced an error.
```
Traceback (most recent call last):
  File "Test/test_python_forecast_coordinates.py", line 64, in setUp
    table_id = cmor.load_table(mip_table)
  File "/Users/mauzey1/opt/anaconda3/envs/cmor_dev/lib/python3.8/site-packages/CMOR-3.6.1-py3.8-macosx-10.9-x86_64.egg/cmor/pywrapper.py", line 916, in load_table
    return _cmor.load_table(table)
_cmor.CMORError: Problem with 'cmor.load_table'. Please check the logfile (if defined).
```

piotr-florek-mohc:
Thanks @mauzey1, I missed the fact that I hadn't put my test CV under version control...

piotr-florek-mohc:
Hi @mauzey1 would you be able to briefly take a look and check how sensible the solution is, and if it requires major reworking? Many thanks!

mauzey1 (approved):

mauzey1:
@piotr-florek-mohc 

I think these changes are reasonable.  However, I would like to have a new entry in the documentation for the `calculate_leadtime_coord` function.  Please create a documentation entry similar to the ones on the [CMOR API doc page](https://cmor.llnl.gov/mydoc_cmor3_api/).  If you could, please create a pull request to add the change using the [doc's repo](https://github.com/PCMDI/cmor3_documentation).  Thank you.

@durack1 @taylor13 Are you fine with these changes?  Do you want me to perform any tests before merging?

durack1:
@taylor13 over to you..

taylor13:
@piotr-florek-mohc - I wonder if you have tested this version as part of your regular job stream in post-processing CMIP6 output (including cases when the new parts of the code are not needed).   If not, do you think such a test is needed to be safe, or are you confident that the coding you've included can't disrupt the older code.  (I know you settled on an implementation strategy that would minimize possible impact, so perhaps the answer is yes, but please confirm.)   Thanks.

mauzey1:
@taylor13 All of the current tests are passing, and the `leadtime` and `reftime` axes only appear in the file for the newly-added test.  I think it's safe to merge.

taylor13:
That's great.  I'm happy.

PR #639, opened 2021-11-22
Reduce length of error message when value cannot be validated

### LINKED ISSUE #638: Illegal instruction error when region attribute is not found in the CV
I encountered an error when working on #628.  It was when I tried to set the `region` attribute for the input4MIPs variable to a string made from a list that I got the "illegal instruction" error.  It turned out that CMOR was trying to create an error message that said that the string used for the attribute value was not found in a regular expression composed of all the [regions listed in the input4MIPs CV file](https://github.com/PCMDI/input4MIPs-cmor-tables/blob/41a6c98858a3c3944fe91139b3bac27e2c336179/Tables/input4MIPs_CV.json#L149-L218).  This causes a problem since there are enough regions in that list to make the regular expression string exceed the maximum string length of CMOR.  The code that produces this message is below.
https://github.com/PCMDI/cmor/blob/e38ba87c81897cef00b765b8f451b250a0855f3a/Src/cmor_CV.c#L2008-L2031

This issue is related to the bigger issue of the maximum string length being only 1024 as listed in #530.  However, rather than tackle that larger issue here, I will just find a solution to handle errors for not finding attribute values in large lists in the CV.  Maybe just mention which attribute in the CV where it couldn't find the values.

mauzey1:
Resolves #638 

This replaces the lengthy error message containing a regular expression string with a smaller message that only lists the section of the CV where the valid values for the attribute are located.  This is to help prevent the message from going over the maximum string length.

Previous message:
```
C Traceback:
! In function: _CV_ValidateAttribute
! called from: _CV_checkGblAttributes
! 

!!!!!!!!!!!!!!!!!!!!!!!!!
!
! Error: The attribute "nominal_resolution" could not be validated. 
! The current input value is "335 km" which is not valid 
! Valid values must match the regular expression:
!       ["^0.5 km$" "^1 km$" "^10 km$" "^100 km$" "^1000 km$" "^10000 km$" "^1x1 degree$" "^2.5 km$" "^25 km$" "^250 km$" "^2500 km$" "^5 km$" "^50 km$" "^500 km$" "^5000 km$" ] 
! 
! Check your Control Vocabulary file "Tables/CMIP6_CV.json".
! 
!
!!!!!!!!!!!!!!!!!!!!!!!!!
```

New message:
```
C Traceback:
! In function: _CV_ValidateAttribute
! called from: _CV_checkGblAttributes
! 

!!!!!!!!!!!!!!!!!!!!!!!!!
!
! Error: The attribute "nominal_resolution" could not be validated. 
! The current input value is "335 km", which is not valid. 
! 
! Valid values must match those found in the "nominal_resolution" section
! of your Control Vocabulary file "Tables/CMIP6_CV.json".
! 
!
!!!!!!!!!!!!!!!!!!!!!!!!!
```

### DIFF
--- a/Src/cmor_CV.c
+++ b/Src/cmor_CV.c
@@ -2019,11 +2019,10 @@ int cmor_CV_ValidateAttribute(cmor_CV_def_t * CV, char *szKey)
         snprintf(msg, CMOR_MAX_STRING,
                  "The attribute \"%s\" could not be validated. \n! "
                  "The current input value is "
-                 "\"%s\" which is not valid \n! "
-                 "Valid values must match the regular expression:"
-                 "\n! \t[%s] \n! \n! "
-                 "Check your Control Vocabulary file \"%s\".\n! ",
-                 szKey, szValue, szValids, CV_Filename);
+                 "\"%s\", which is not valid. \n! \n! "
+                 "Valid values must match those found in the \"%s\" "
+                 "section\n! of your Control Vocabulary file \"%s\".\n! ",
+                 szKey, szValue, szKey, CV_Filename);
 
         cmor_handle_error(msg, CMOR_NORMAL);
         cmor_pop_traceback();
--- a/Src/cmor_CV.c
+++ b/Src/cmor_CV.c
@@ -2009,19 +2009,12 @@ int cmor_CV_ValidateAttribute(cmor_CV_def_t * CV, char *szKey)
 /* We could not validate this attribute, exit.                          */
 /* -------------------------------------------------------------------- */
     if (i == (attr_CV->anElements)) {
-        for (i = 0; i < attr_CV->anElements; i++) {
-            strcat(szValids, "\"");
-            strncpy(szOutput, attr_CV->aszValue[i], CMOR_MAX_STRING);
-            strcat(szValids, szOutput);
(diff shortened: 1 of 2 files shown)


Src/cmor_CV.c
```diff
@@ -2019,11 +2019,10 @@ int cmor_CV_ValidateAttribute(cmor_CV_def_t * CV, char *szKey)
...
                  "The attribute \"%s\" could not be validated. \n! "
                  "The current input value is "
-                 "\"%s\" which is not valid \n! "
-                 "Valid values must match the regular expression:"
-                 "\n! \t[%s] \n! \n! "
-                 "Check your Control Vocabulary file \"%s\".\n! ",
-                 szKey, szValue, szValids, CV_Filename);
+                 "\"%s\", which is not valid. \n! \n! "
+                 "Valid values must match those found in the \"%s\" "
+                 "section\n! of your Control Vocabulary file \"%s\".\n! ",
```
durack1, line 2024: @mauzey1 updating `..Control Vocabulary..` to `..Controlled Vocabulary (CV)..` would be a good idea

durack1 (approved):
@mauzey1 approved with the suggested edit

PR #641, opened 2021-12-13
remove unused part of the interface

piotr-florek-mohc:
I removed two leftovers from the earlier part of the development, when one would need to perform an explicit call to the function creating the leadtime coordinate. Because it's all internalised now, there is no need for exposing that part of the CMOR code, and provide Python (or Fortran) API.

### DIFF
--- a/Src/_cmormodule.c
+++ b/Src/_cmormodule.c
@@ -452,23 +452,6 @@ static PyObject *PyCMOR_load_table(PyObject * self, PyObject * args)
 }
 
 /************************************************************************/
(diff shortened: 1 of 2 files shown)


mauzey1:
@piotr-florek-mohc Thank you.  So this should make the mention of this function unnecessary in the API documentation.  Should I decline that pull request?

mauzey1 (approved):

piotr-florek-mohc:
> @piotr-florek-mohc Thank you. So this should make the mention of this function unnecessary in the API documentation. Should I decline that pull request?

Yes please! Sorry about the confusion.

PR #655, opened 2022-05-11
Properly remove multiple comment sections from cell_methods values

### LINKED ISSUE #654: PrePARE is having issues handling handling cell_methods values containing multiple comments
The values for the `cell_methods` can contain [comments within parentheses](https://github.com/PCMDI/cmip6-cmor-tables/blob/cf039eb222cb0b7b92b3df38d986b8015450a828/Tables/CMIP6_Omon.json#L1840).
```
 "cell_methods": "longitude: sum (comment: basin sum [along zig-zag grid path]) depth: sum time: mean", 
```
However, some of the files generated by CMOR can create values with multiple parentheses sections.
```
htovgyre:cell_methods = "longitude: sum (comment: basin sum [along zig-zag grid path]) depth: sum time: mean (interval: 1 month)"
```
PrePARE is currently only designed to handle removing one parentheses section.  Having multiple sections will cause everything from the first starting parenthesis to the last ending parenthesis to be removed.  This will then cause an error when comparing the values between the CMIP6 table and the dataset.
```
=====================================================================================
Your file contains "cell_methods":"longitude: sum" and
CMIP6 tables requires "cell_methods":"longitude: sum depth: sum time: mean".
=====================================================================================
```

I have made a change to the regex that handles removing the comment sections correctly.  I will create a PR for it shortly.

mauzey1:
Fixes #654 

The regex will now only remove multiple comment sections rather everything between them.
```
"longitude: sum (comment: basin sum [along zig-zag grid path]) depth: sum time: mean (interval: 1 month)"
```
becomes...
```
"longitude: sum  depth: sum time: mean "
```
instead of...
```
"longitude: sum "
```

### DIFF
--- a/LibCV/PrePARE/PrePARE.py
+++ b/LibCV/PrePARE/PrePARE.py
@@ -617,8 +617,8 @@ class checkCMIP6(object):
                         table_value = file_value
                 if key == "cell_methods":
                     # Remove text that is inside parentheses i.e. comments
-                    file_value = re.sub(r"\(.*\)", "", file_value)
-                    table_value = re.sub(r"\(.*\)", "", table_value)
+                    file_value = re.sub(r"\(.*?\)", "", file_value)
+                    table_value = re.sub(r"\(.*?\)", "", table_value)
                     # Remove extra whitespace
                     file_value = " ".join(file_value.split())
                     table_value = " ".join(table_value.split())


durack1 (commented):
@mauzey1 thanks for catching this

durack1 (approved):

taylor13:
Yes, thanks Chris.  By the way,  which table and variable did the "double comment" you used as an example above come from?

taylor13:
Thanks for that, Chris. I wasn't aware that CMOR adds information about "interval" to what it reads from the CMOR table.  I couldn't tell from the coding whether the "interval" gets added for any "time" dimension in the ``cell_methods`` attribute, or for just a few special cases.  I don't think CMOR should ever modify in any way the ``cell_methods`` it reads from the CMOR table.  (Let's review this further before modifying the code though.)

Note also, that the interval of "1 month" that CMOR generates is almost surely incorrect.  The "interval" is defined [here](https://cfconventions.org/cf-conventions/cf-conventions.html#recording-spacing-original-data) in the CF conventions.  It is supposed to indicate the _**original**_ sampling interval used in compiling a mean (or other statistic).  Many time-means are computed from model data sampled every time step, and the time step will vary from one model to another, so CMOR can't know what the correct value of the "interval" should be, and thus should not attempt to define the interval.

Any thoughts?  Does this make sense?

taylor13:
By the way, your new coding is still needed as there may be future data requests where more than one comment rightfully appears in the ``cell_methods``, so thanks for that.

PR #661, opened 2022-08-15
Remove numpy.distutils.core and unnecessary modules from setup.py.in

mauzey1:


durack1 (commented):
@mauzey1 these were presumably hangovers from past work depending on `cdms2` right?

mauzey1:
@durack1 I'm not sure why those modules were included in setup.py, but they were causing a conda build to fail and were not essential to the build process.

durack1:
@mauzey1 thanks for the continuing cleanup - glad you're catching these remaining issues

PR #672, opened 2022-08-26
Fix PrePARE for Python 3.10 builds

### LINKED ISSUE #670: PrePARE skipping files in Python 3.10 builds
When testing PrePARE for Python 3.10, I get the following message for a valid file.

```
$ PrePARE rlut_Amon_UKESM1-0-LL_histSST-piAer_r1i1p1f2_gn_195001-201412.nc 
<built-in function check_filename> returned a result with an exception set
└──> :: SKIPPED    :: /Users/mauzey1/Desktop/github/cmor/rlut_Amon_UKESM1-0-LL_histSST-piAer_r1i1p1f2_gn_195001-201412.nc

Number of files scanned: 1
Number of file with error(s): 1
```
What I get with the Python 3.9 version...
```
$ PrePARE rlut_Amon_UKESM1-0-LL_histSST-piAer_r1i1p1f2_gn_195001-201412.nc 

Number of files scanned: 1
Number of file with error(s): 0
```
This is an issue that might affect newer versions of Python so we should try fixing this issue before adding a Python 3.11 build (#669).

mauzey1:
Fixes #670 

I discovered that PrePARE was failing at the check_filename function of Python 3.10 builds.  It was due to an error occurring with a PyArg_ParseTuple call.
https://github.com/PCMDI/cmor/blob/7cb3f2a0ea22defe24019626c5260e6514678c21/Src/_controlvocabulary.c#L13-L37
The PyArg_ParseTuple call was using the argument format `iis#s#s#`, where the `s#` was meant to both get the value of a string and its length.  This was not working correctly in Python 3.10 as the string and length variables were not being initialized with the argument values.

It turned out that using the argument format of `iisss` didn't have this issue.  This means that we get the string values without their lengths.  However, the lengths of the strings were not necessary for the file checking function so the change had no impact to the rest of the function.

### DIFF
--- a/LibCV/pywrapper.py
+++ b/LibCV/pywrapper.py
@@ -270,18 +270,21 @@ def check_sourceID(table_id):
     return(ierr)
 
 
-def check_filename(table_id, var_name, calendar, timeunits, infile):
+def check_filename(table_id, var_id, calendar, timeunits, infile):
     '''
       Validate filename with timestamp for current variable and file
 
       Usage:
-        cmip6_cv.check_filename(table_id, var_id)
+        cmip6_cv.check_filename(table_id, var_id, calendar, timeunits, infile)
       Where:
         table_id is the table id returned by load_table()
-        var_name is the variable name
+        var_id is the variable id returned by setup_variable()
+        calendar is the calendar type used by the file (ex. 360_day)
+        timeunits is the time unit type used by the file (ex. days since 1850-01-01)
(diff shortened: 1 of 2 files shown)


durack1 (approved):
@mauzey1 I am unfamiliar with C, so will defer to your judgement on this - I think we'll need to keep and eye on this as Python 3.11 is released

PR #680, opened 2022-11-01
Multiple write bug

piotr-florek-mohc:
This is is a fix to a bug that causes CMOR failure when doing multiple writes to the dataset with seasonal forecast time coordinates ('reftime' and 'leadtime'). The original implementation didn't take into account a fact that a coordinate definition could happen only once and the code would fail during a second write() call.


mauzey1 (approved):

mauzey1:
@piotr-florek-mohc This looks good, but I will need to fix the python 3.7 tests before merging.

PR #696, opened 2023-02-23
Merge CMIP6 CV tests into Makefile test section

mauzey1:
This is to help simplify the testing process for source builds.  `make test` will now also run the CMIP6 CV Python tests that were once ran using testsrunner.


durack1 (approved):
@mauzey1 thanks!

PR #701, opened 2023-06-02
Throw error if the "further_info_url" in the user input is an empty string or whitespace

### LINKED ISSUE #700: CMOR segfaults when further_info_url is specified as an empty or pure whitespace string
I've found a minor issue that has been causing me a few issues.

If you specify the `further_info_url` string in the JSON file provided to `cmor.dataset_json()` and you provide a string that does not match the pattern in the CVs you'll get an appropriate error when `cmor.write()` is called;
```
! Error: The attribute "further_info_url" could not be validated. 
! The current input value is "incorrect_string", which is not valid. 
```

If you omit a `further_info_url` entry from the dataset JSON file altogether then CMOR will construct this field for you appropriately and everything works fine.

However, if you include a `further_info_url` entry which is either an empty string or only contains whitespace then CMOR segfaults when `cmor.write()` is called, giving no information on what is going on. 

Would it be possible to augment the validation of this field?

mauzey1:
Resolves #700

This change will make CMOR throw an error message if you set `further_info_url` in the user input file to be an empty string or only whitespace.


durack1 (approved):
@mauzey1 looks good!

PR #708, opened 2023-09-22
706 writing large volume of data

### LINKED ISSUE #706: CMOR failure when attempting to write large volume of data

mauzey1:
Resolves #706 

This PR will replace the use of `int` with `size_t` in the function `cmor_write_var_to_file` when calculating the number of elements in an array.  This will allow `cmor_write` to write data arrays with more than 2^31-1 elements.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -98,7 +98,7 @@ executors:
         CONDA_COMPILERS: "gcc_linux-64 gfortran_linux-64"
    macos:
       macos:
-         xcode: "13.2.1"
+         xcode: "13.4.1"
       environment:
         OS: "osx-64"
         PROJECT_DIR: "workdir/macos"
--- a/Src/cmor_variables.c
+++ b/Src/cmor_variables.c
@@ -2344,32 +2344,32 @@ int cmor_write_var_to_file(int ncid, cmor_var_t * avar, void *data,
 
     size_t counts[CMOR_MAX_DIMENSIONS];
     size_t counts2[CMOR_MAX_DIMENSIONS];
-    int counter[CMOR_MAX_DIMENSIONS];
-    int counter_orig[CMOR_MAX_DIMENSIONS];
-    int counter_orig2[CMOR_MAX_DIMENSIONS];
-    int counter2[CMOR_MAX_DIMENSIONS];
+    size_t counter[CMOR_MAX_DIMENSIONS];
+    size_t counter_orig[CMOR_MAX_DIMENSIONS];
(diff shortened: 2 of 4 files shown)


durack1 (approved):
@mauzey1 this is great, thanks!

There maybe some very subtle changes required to deal with the new MIP-cmor-tables, a string -> list type change for coord descriptions, so then we can wrap that up as 3.7.3 or 3.8 if the osx-arm64 change happens in the same release 

Ping @matthew-mizielinski @taylor13

durack1:
We'll need to check on those Linux build failures too

mauzey1:
> There maybe some very subtle changes required to deal with the new MIP-cmor-tables, a string -> list type change for coord descriptions

Could you elaborate on this feature?  Please post an issue detailing it.

> We'll need to check on those Linux build failures too

The conda rerender step is taking too long when installing packages for some reason.  I encountered that issue when working on another branch but I was able to resolve it by just rerunning the failed workflows.  I will try rerunning them after the OSX builds are done.

durack1:
@mauzey1 will do (post a new issue). I am travelling next week and will meet with Matt in person, so we can figure out what is required and document that in an issue.

And yep, you're right, second pass on circleci passes fine

Issue #710, opened 2023-09-26
"call to undeclared function 'calculate_leadtime_coord'" error in recent Xcode/Clang build for OSX


mauzey1:
While building CMOR on my Macbook I encountered the following error.
```
Src/cmor.c:4859:9: error: call to undeclared function 'calculate_leadtime_coord'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
        calculate_leadtime_coord(var_id);
```
I'm currently using Xcode 15.0 on OSX Ventura.  I've been building CMOR on my Mac just fine until recently.  This issue might have arisen after I rebuilt my Conda environment with the latest version of Clang.

Regardless, the error is caused by the call to the function `calculate_leadtime_coord` in [cmor.c](https://github.com/PCMDI/cmor/blob/292efd9bd23ad748cff91a8919656f92e857db87/Src/cmor.c#L4859) before its [declaration](https://github.com/PCMDI/cmor/blob/292efd9bd23ad748cff91a8919656f92e857db87/Src/cmor.c#L6646).  The simplest solution would be to just declare the function near the [top of the file](https://github.com/PCMDI/cmor/blob/292efd9bd23ad748cff91a8919656f92e857db87/Src/cmor.c#L45).

matthew-mizielinski:
@piotr-florek-mohc, I'm guessing that you didn't see this when doing the leadtime work?

doskonaleszare:
no I haven't - seems like an embarrassingly rookie mistake!

piotr-florek-mohc:
no I haven't - seems like an embarrassingly rookie mistake!

mauzey1:
I never caught it either until the changes to the compiler flagged it as an error.

@piotr-florek-mohc, if you want to make some fixes to this I could hand this issue over to you.  I also noticed there were some functions in the [Python wrapper ](https://github.com/PCMDI/cmor/blob/292efd9bd23ad748cff91a8919656f92e857db87/Lib/pywrapper.py#L461) that were meant to point this function until it was decided not to expose it in the API.

piotr-florek-mohc:
Sure, will fix both!

PR #713, opened 2023-10-04
issue 710: fixing implicit declarations and updating Python API

piotr-florek-mohc:
Fixes for #710 and removal of a leftover Python function (as there is no need to insert the leadtime coordinate manually).


mauzey1:
@piotr-florek-mohc There's also an import of `calculate_leadtime_coord` in Lib/\_\_init\_\_.py you'll need to remove.
https://github.com/PCMDI/cmor/blob/292efd9bd23ad748cff91a8919656f92e857db87/Lib/__init__.py#L16

mauzey1 (approved):

PR #717, opened 2023-10-21
Speed up rerender for nightly builds

mauzey1:
The rerendering step for the nightly builds have been really slow.  They have been taking up to 30 minutes to complete.  The issue was caused by trying to install the conda build packages into the base miniconda environment.  Installing to the base environment made conda process package conflicts, which can take a long time and will sometimes cause the build to fail if it takes too long.

The solution was to have the rerender step create another conda environment with the conda build packages.  This has reduced the time of the rerender step from around 30 minutes to around 1 minute.


durack1:
Great catch, thanks for sorting this @mauzey1!

taylor13:
Agreed.  Very nicely identified and solved!

PR #727, opened 2024-02-16
Make further_info_url attribute only be used if required by CV or requested by user

mauzey1:
This PR will make CMOR only add a `further_info_url` attribute to the file if it is either required by the CV, or set in the user input file.  If `further_info_url` is in the `required_global_attributes` section of the CV, it will require CMOR to add that attribute to the file.

The `further_info_url` attribute value provided by the user in the input file is a template that will be used by CMOR to generate a URL.  If the user doesn't provide a template value when it is required, then CMOR will use its built-in default template for the URL.  If the CV file provides a value for `further_info_url` in the `source_id` entry, then it will be used as a value for the file's attribute unless `further_info_url` is set in the user input.

### DIFF
--- a/Src/cmor.c
+++ b/Src/cmor.c
@@ -1321,7 +1321,7 @@ int cmor_dataset_json(char *ressource)
                 CMOR_DEFAULT_FILE_TEMPLATE, CMOR_MAX_STRING);
 
     strncpytrim(cmor_current_dataset.furtherinfourl,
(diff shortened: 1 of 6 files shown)


piotr-florek-mohc (approved):
Looking good to me, and the code behaves as expected, with CMOR complaining about missing attribute if I put further_info_url back into CVs.

Issue #601, opened 2020-05-06
Performance issues with small chunks


cofinoa:
We are facing some performance issue accessing to metadata, i.e. values for time variable, because the number of I/O reading operations required to access all the chunks. 

In particular the time coordinate variable it's created with chunk size 1, requiring one chunk per time value, therefore if in the netcdf-4 there is a lot of time steps (in 6-hr or 3-hr > 10k), the netcdf-4 library has too look and read for each chunk (i.e. 8 bytes per chunk). 

A better explanation of this pitfall, can be found on [1]: 
>Chunks are too small
There is a certain amount of overhead associated with finding chunks. When chunks are made smaller, there are more of them in the dataset. When performing I/O on a dataset, if there are many chunks in the selection, it will take extra time to look up each chunk. In addition, since the chunks are stored independently, more chunks results in more I/O operations, further compounding the issue. The extra metadata needed to locate the chunks also causes the file size to increase as chunks are made smaller. Making chunks larger results in fewer chunk lookups, smaller file size, and fewer I/O operations in most cases.

This relates to: #99, #100, #164

[1] https://support.hdfgroup.org/HDF5/doc/Advanced/Chunking/

taylor13:
@cofinoa You indicated in one of the related postings that in netCDF3 making larger chunks for the time coordinate means that it can't be declared "unlimited".  In netCDF4 is that also true or can it be declared "unlimited" and be made into bigger chunks?
Thanks.

durack1:
@mauzey1 is there a preset chunking value set in the code somewhere? I recall going over this in some detail many years ago, but a quick search of the repo for "chunk" doesn't appear to have shown any defaults, at least in my viewing

cofinoa:
@taylor13 to mitigate the problem in netcdf-3 the only solution it's not to make unlimited time dimension.

In netcdf4/hdf5 you can select different chunksizes to make it bigger size for the time coordinate variable and chunksize of 1 for the principal variable.

cofinoa:
@durack1 and @mauzey1 the PR #100 just merge a change to impose a chunking size of 1 to time coordinate. 

https://github.com/PCMDI/cmor/pull/100/commits/fc738dfa35e978721ae8ebb70a3627b82861961f

taylor13:
[...]

PR #733, opened 2024-04-09
Update cmor.c

cofinoa:
Use netcCDF4 DEFAULT_CHUNK_SIZES, for chunked vars and coordinates/axis.

This relates to issue #601 where it is explained that chunk sizes of 1, for coordinates/axis, like `time` has a huge bad performance impact on reading those netCDF variables. 

The netcdf-c library defines default CHUNK sizes for netCDF4/HDF5 files when chunkingsizes are NULL. 

For current netcdf-c (i.e. version 4.9.2) 

 * `nc_def_var_chunking`: 
   > [...] Chunk sizes may be specified with the chunksizes parameter or default sizes will be used if that parameter is NULL. [...]
 * See [Default Chunking Scheme](https://docs.unidata.ucar.edu/nug/current/netcdf_perf_chunking.html#default_chunking_4_1) from netCDF User Guide (NUG): 
   * > [...] variables that only have a single unlimited dimension [...] the [default] chunk sizes for such variables are limited to 4KiB
   * > [...] Currently the netCDF default chunk size is 4MiB, which is reasonable for filesystems on high-performance computing platforms [...]
   * > [...] The current default chunking strategy of the netCDF library is to balance access time along any of a variable's dimensions, by using chunk shapes similar to the shape of the entire variable but small enough that the resulting chunk size is less than or equal to the default chunk size. This differs from an earlier default chunking strategy that always used one for the length of a chunk along any unlimited dimension, and otherwise divided up the number of chunks along fixed dimensions to keep chunk sizes less than or equal to the default chunk size. [...]
 * To change the default chunk cache size, use the `nc_set_chunk_cache()` function before opening the file, for all variables, or per variable use `nc_set_var_chunk_cache()`.
 * Related HDF5 function: [H5Pset_cache](https://support.hdfgroup.org/HDF5/doc/RM/RM_H5P.html#Property-SetCache)
 * This PR not only propose DEFAULT chunking for time coordinate/axis but also for data variable itself with unlimited dimensions.

### DIFF
--- a/Src/cmor.c
+++ b/Src/cmor.c
@@ -3588,7 +3588,7 @@ void cmor_define_dimensions(int var_id, int ncid,
                 && (CMOR_NETCDF_MODE != CMOR_APPEND_3)) {
                 if (strcmp(pAxis->id, "time") == 0) {
                     ierr = nc_def_var_chunking(ncid, nc_vars[i], NC_CHUNKED,
-                                               &nc_dim_chunking[0]);
+                                               NULL);
                 } else {
                     ierr = nc_def_var_chunking(ncid, nc_vars[i], NC_CONTIGUOUS,
                                                &nc_dim_chunking[0]);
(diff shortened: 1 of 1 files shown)


mauzey1 (approved):
The changes look good.  Please merge the latest changes from main to this branch.

durack1:
@cofinoa it seems like you need to rebase your branch with the latest `main` and then repush? We may need to stand up a new test alongside your PR changes

PR #741, opened 2024-05-03
Remove default value for attribute 'realm'

### LINKED ISSUE #736: default `realm = "REALM"` is always written although not required by CV
I think this is a similar issue as https://github.com/PCMDI/cmor/pull/727. For CORDEX, we have eliminated the `realm` attribute from the required global attributes (and also from table headers). However, cmor will write a default of `realm = REALM` ([example](https://wcrp-cordex.github.io/cordex-cmip6-cmor-tables/cmor-examples.html#example-from-fx-table)) no matter what.

mauzey1:
Closes #736 

This will remove the setting of the `realm` attribute to the default value of `REALM`.  `realm` will still be set to the `modeling_realm` attribute's value from the variable's table even if `realm` is not a required attribute in the CV.

### DIFF
--- a/Src/cmor_tables.c
+++ b/Src/cmor_tables.c
@@ -83,7 +83,7 @@ void cmor_init_table(cmor_table_t * table, int id)
     table->cmor_version = 3.0;
     table->mip_era[0] = '\0';
     table->szTable_id[0] = '\0';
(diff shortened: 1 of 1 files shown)


durack1 (approved):
Wonderful, thanks @mauzey1

PR #750, opened 2024-07-10
Add GitHub Action build pipeline with Apple M1 builds

### LINKED ISSUE #663: arm64 build
There are now a number of conda variants that are supporting the Apple M1/`arm64` architecture natively. As a low priority item, it would be interesting to investigate how easy/hard native CMOR builds would be to generate and support - does the CI environments support this architect natively yet?

Label CMOR 4.x for this one

mauzey1:
Resolves #663 

This will add a GitHub Actions pipeline will make MacOS builds for both x86 and Apple M1 machines.  The pipeline will only launch jobs whenever a PR is made or updated, and will only deploy nightly builds when merging to main.


durack1 (approved):
@mauzey1 great! It looks like there are some tests that have failed and been cancelled, but I presume this is all ok so we're good to merge

mauzey1:
@durack1 There are still some issues I need to fix so I'll mark this as a draft.

PR #764, opened 2024-11-14
New error function and resolved compiler warnings

### LINKED ISSUE #602: Numerous warnings when compiling CMOR
When CMOR is built using `make`, the C compiler will output many warnings.  A lot of them are related to snprintf, which is due to the destination string potentially being not long enough to store the entire string being generated.
```
Src/cmor_CV.c: In function 'cmor_CV_checkExperiment':
Src/cmor_CV.c:1473:58: warning: '%s' directive output may be truncated writing up to 1023 bytes into a region of size 1002 [-Wformat-truncation=]
                                  "Your input attribute \"%s\" with value \n! \"%s\" "
                                                          ^~
```
Some are due to comparisons between different data types.
```
Src/cmor.c: In function 'cmor_pop_traceback':
Src/cmor.c:434:19: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
     for (i = 0; i < strlen(cmor_traceback_info); i++) {
                   ^
```
Some are due to function declarations that should be prototypes in the header files.  Functions that don't have parameters should be declared with `(void)` instead of `()`.
```
In file included from ./include/cmor.h:578:0,
                 from Src/cmor_variables.c:4:
./include/cmor_func_def.h:15:1: warning: function declaration isn't a prototype [-Wstrict-prototypes]
 extern void cmor_set_terminate_signal_to_sigint();
 ^~~~~~
```
Although we could suppress these warning messages, it would be better to fix these issues in the code.

mauzey1:
Resolves #602

This PR resolves numerous compiler warnings that spring up when building CMOR or its C tests.

Part of this effort involved creating a new error handling function that uses variadic arguments for format strings similar to `printf`.  This will allow for error messages to have strings greater than 1023 characters (`CMOR_MAX_STRING` - 1 for null-terminated strings).


durack1 (approved):
@mauzey1 this is great, thanks for pushing on this!

Issue #762, opened 2024-10-04
CMIP7 requirements: "branded variable" and new mip_table specification


taylor13:
(FYI @sashakames, @durack1,@matthew-mizielinski, @wolfiex  even though this is primarily for Chris)

It looks likely that some changes to the output requirements for CMIP7 will be agreed shortly and that "branded variables" will be relied on in identifying variables in the cmor output files.  It would be good to now consider how this might impact CMOR, so I'll raise this issue now:

How difficult would it be to implement the following? 

1. The user specifies “frequency” as one of the entries in the CMIP6_input.json file rather than it being specified in a CMOR variable table.   CMOR then handles “frequency” in the same way it handles, for example, “experiment_id”, and writes it as a global attribute.  (We would also remove “frequency” and “approx_interval” from the CMOR variable tables.)  I know that CMOR checks that users have sent a time coordinate that is approximately consistent with "approx_frequency", but that check could be dropped if it impairs implementation of this new approach.
2. The user specifies “region” as one of the entries in the CMIP6_input.json file and then CMOR handles it in the same way as, for example, “experiment_id” and writes it as a global attribute?
3. CMOR writes as a global attribute the "branding suffix", which it would need to obtain by extracting the suffix in the "table_entry" (i.e., the part following the underscore).  See below for an example.
4. CMOR writes as global attributes the values of the elements comprising the branding suffix: temporal_sampling, vertical_sampling, horizontal_sampling, and area_sampling?  These would be either be extracted, along with other metadata, from the CMOR variable table (as shown in the table example below), or could be obtained from a look-up table given the branding suffix.
5. In constructing file names and directory structure, rely on a somewhat different set of global attributes than in CMIP6.  For example instead of including “table name” in the file name, include instead the “branded variable suffix”.  (My guess is that this is trivially done by simply specifying a different template in the CMIP6_input.json file.)

To implement the above, new CMOR variable tables will need to be generated with the following changes (which could be implemented by someone other than Chris):

1. Remove "approx_interval" from the header of each table.
2. Remove “frequency” from each entry in the variable tables.
3. Replace all the variable “table_entries” with branded variable names.
4. Make sure out_name is set to the root name prefix of the branded variable (i.e., the part of the branded variable name preceding the underscore). 
5. Add 5 new attributes to each variable in the tables: branding_suffix, temporal_type, vertical_type, horizontal_type, and area_type.  These will be written by CMOR as global attributes in the netCDF files. 
[...]

PR #778, opened 2025-02-26
Add placeholder CMIP7 MIP tables, and CMIP7_CV.json (#762)

durack1:
Ref #762 

@mauzey1 @taylor13 @matthew-mizielinski @wolfiex ping - first pass at placeholder template files


mauzey1:
Will the variable tables in this pull request eventually have the branded variable format?  Will those changes be reflected in mip-cmor-tables?

taylor13:
I'll be happy to edit the python code that creates these tables to be consistent with the table structure described in https://github.com/PCMDI/cmor/issues/762 if you point me to where it is.

mauzey1:
Aside from needing `CMIP7_coordinate.json` and `CMIP7_formula_terms.json`, there should be the following sections in `CMIP7_CV.json`.
- `required_global_attributes`
- `institution_id`
- `experiment_id`
- `sub_experiment_id`
- `source_id`.

durack1:
@mauzey1 great, I'll pull some placeholder entries across so we're starting to build out a file for CMIP7/CMOR3.10 testing - expect an updated commit to drop later today

taylor13:
regarding https://github.com/PCMDI/cmor/pull/778#issuecomment-2695654907, do we really expect the structure of the coordinates, formulas, and CMIP_CV.json file to be different from CMIP6?  I think only if we need to associate an approx_interval with each frequency, but not if that's hard-wired within CMOR.    Everything else should be the same as CMIP6, right?  There will certainly be changes in content, but we have to wait for that.

durack1:
@taylor13 I think you're right, the contents of `CMIP6_coordinates.json`, `*formula_terms.json`, and `*grids.json` are not likely to need changing at this moment, so these can be used directly from CMIP6/CMIP6Plus or any other project that has duplicated the contents.

The current version of this PR now includes the changes I believe are captured in #762 - might have some final tweaks, but these are almost ready to start working with

TestTables/CMIP7_CV.json
```diff
@@ -0,0 +1,206 @@
...
+        },
+        "required_global_attributes":[
+            "Conventions",
+            "activity_id",
+            "creation_date",
+            "data_specs_version",
+            "experiment",
+            "experiment_id",
+            "forcing_index",
+            "frequency",
```
taylor13, line 113: In addition to 
"frequency", we now need:
"region"
"branding_suffix"
"temporal_label"
"vertical_label"
"horizontal_label"
"area_label"
"host_collection"
"archive_id"

what about parent_experiment_id and related attributes?  

I'm concerned that we have no "branching" information.  In CMIP6, at least 7 global attributes recorded this information.  How will we get groups to record this stuff if not in the file headers?

TestTables/CMIP7_CV.json
```diff
@@ -0,0 +1,206 @@
...
+            "nominal_resolution",
+            "physics_index",
+            "realization_index",
+            "realm",
+            "source",
+            "source_id",
+            "source_type",
+            "sub_experiment",
+            "sub_experiment_id",
+            "table_id",
```
taylor13, line 130: not sure the users will care about table_id.  probably should delete it.

taylor13 (commented):
Lots of changes needed as indicated.  A couple may deserve some discussion.

TestTables/CMIP7_CV.json
durack1: What is the source of these CVs, particularly region, branding_suffix, *_label, host-collection, archive_id?

taylor13: none of them existed in CMIP6, but they will be put in place for CMIP7.  The branding_suffix and *_label CV's will get their information from Appendix F of the draft journal article.
archive_id = "WCRP"
which distinguishes the MIP projects from other projects like E3SM and distinguishes one governance body/vocabularies/name spaces/formats/data structures/ from a different set of these.  This should be specified in the CV.json file needed by CMOR.

The regions needed for CMIP7 are "global", "antarctica", "greenland", and possibly "northern_hemisphere" and "southern_hemisphere" (will need to check with the CMIP7 data request).  We'll have a "region" CV that should get incorporated into the CMIP7_CV.json file.  Along with frequency, the user will set the value for a given output dataset.  Chris is aware of this.  [For region, we might set a default value of "global" in the CMIP7_CV.json file, since that applies to almost all model output.]

the branding_suffix and *_labels will be extracted from the variable entry (i.e., the full name of the branded variable) by a simple "parsing" procedure.  There is no need to check these against a controlled vocabulary since all entries in the CMOR tables will have legal branded variable names (and suffixes).   We want to store the elements comprising the branded variable name as separate global attributes so they can easily used by ESGF as search facets.

durack1: We have "branch_method", "branch_time_in_child", "branch_time_in_parent", in the global att doc [here](https://docs.google.com/document/d/1h0r8RZr_f3-8egBMMh7aqLwy3snpD6_MrDz1q8n5XUk/edit), are you talking about the "parent_*" attributes too?

durack1: Ok and adding to the list: "parent_activity_id", "parent_experiment_id", "parent_mip_era", "parent_source_id", "parent_time_units", "parent_variant_label",

durack1: And added *_label elements from the appendices in the "Defining and naming variables.." doc [here](https://docs.google.com/document/d/19jzecgymgiiEsTDzaaqeLP6pTvLT-NzCMaq-wu-QoOc/edit)

TestTables/CMIP7_atmos2d.json
```diff
@@ -0,0 +1,71 @@
...
+            "cell_methods":"area: time: mean",
+            "comment":"includes both liquid and solid phases",
+            "dimensions":[
+                "longitude",
+                "latitude",
+                "time"
+            ],
+            "frequency":"day",
+            "long_name":"Precipitation",
+            "modeling_realm":[
```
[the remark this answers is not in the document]
durack1: There may be case with more than a single realm, no..?

TestTables/CMIP7_atmos2d.json
```diff
@@ -0,0 +1,71 @@
...
+                "atmos"
+            ],
+            "ok_max_mean_abs":"",
+            "ok_min_mean_abs":"",
```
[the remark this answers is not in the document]
durack1: Didn't we want to keep this, as propagating QC range checking info from CMIP6?

TestTables/CMIP7_atmos2d.json
```diff
@@ -0,0 +1,71 @@
...
+            "out_name":"pr",
+            "positive":"",
```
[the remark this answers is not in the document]
durack1: Ok I get it, if the entry is blank, then it's redundant within the variable definition. Will do the same for `ok_max_mean_abs` and `ok_min_mean_abs`.

What does "positive" relate to, this is the sign, so "up" or "down" are valid entries, right?

TestTables/CMIP7_ocean2d.json
```diff
@@ -0,0 +1,46 @@
...
+            ],
+            "ok_max_mean_abs":"",
+            "ok_min_mean_abs":"",
+            "out_name":"tos",
+            "positive":"",
+            "standard_name":"sea_surface_temperature",
+            "type":"real",
+            "units":"degC",
+            "valid_max":400.0,
+            "valid_min":300.0
```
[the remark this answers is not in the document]
durack1: Units degC so valid max = 100, valid_min - -50

durack1:
@taylor13 ok this is where I am up to - this is getting very close. Feel free to nit on what I have in the file changed in this PR

TestTables/CMIP7_CV.json
```diff
@@ -0,0 +1,295 @@
+{
+    "CV":{
+        "area_label":[
+            "air",
```
durack1, line 4: I could see utility in converting this into a dictionary, adding the descriptive text
```python
"air":"air",
...
"gis":"grounded_ice_sheet",
...
"lus":"sector (auxiliary coordinate for land-use area types)"
...
```

Doing this across all labels would be useful

TestTables/CMIP7_atmos2d.json
```diff
@@ -0,0 +1,61 @@
...
+        "int_missing_value":"-999",
+        "missing_value":"1e20",
+        "ok_max_mean_abs":"",
+        "ok_min_mean_abs":"",
+        "positive":"",
+        "product":"model-output",
+        "realm":"atmos",
+        "table_date":"2025-03-05",
+        "table_id":"atmos2d",
+        "type":"real",
```
durack1, line 16: `real` is a Fortran data type, if we're cleaning up, should map to modern C/Python types, most often being `float`

durack1: From what I can ascertain, CMOR will only accept `real` (`float32`) and `integer` (`int32`) types, with no ability to more granularly use the space-saving data types of `byte/int8`, `ubyte/uint8`, `short/int16`, and `ushort/uint16` vs the default `real/float32` type

TestTables/CMIP7_CV.json
```diff
@@ -0,0 +1,297 @@
...
+                "approx_interval":0.041666666666666664,
+                "description":"sampled hourly"
+            },
+            "1hrCM":{
+                "approx_interval":0.041666666666666664,
+                "description":"monthly-mean diurnal cycle resolving each day into 1-hour means"
+            },
+            "3hr":{
+                "approx_interval":0.125,
+                "description":"3 hourly mean samples"
```
matthew-mizielinski, line 126: There will be some time: mean and some time: point variables sharing this frequency. Would it be best to replace this with just "3 hourly"?

TestTables/CMIP7_CV.json
durack1: Adding some extra nits, following @taylor13 replies

> CMIP7 data archive_id = "WCRP" which distinguishes the MIP projects from other projects like E3SM and distinguishes one governance body/vocabularies/name spaces/formats/data structures/ from a different set of these.

> The regions needed for CMIP7 are "global", "antarctica", "greenland", and possibly "northern_hemisphere" and "southern_hemisphere" (will need to check with the CMIP7 data request)

durack1:
Adding some extra nits, following @taylor13 replies

> CMIP7 data archive_id = "WCRP" which distinguishes the MIP projects from other projects like E3SM and distinguishes one governance body/vocabularies/name spaces/formats/data structures/ from a different set of these.

> The regions needed for CMIP7 are "global", "antarctica", "greenland", and possibly "northern_hemisphere" and "southern_hemisphere" (will need to check with the CMIP7 data request)

mauzey1:
Another section that is needed in the CV is the `nominal_resolution` section.
```
        "nominal_resolution":[
            "0.5 km",
            "1 km",
            "10 km",
            "100 km",
            "1000 km",
            "10000 km",
            "1x1 degree",
            "2.5 km",
            "25 km",
            "250 km",
            "2500 km",
            "5 km",
            "50 km",
            "500 km",
            "5000 km"
        ],
```

TestTables/CMIP7_CV.json
taylor13: Yes, definitely.

taylor13:
Hi Matt,
some responses:

>     I'm still a little uncomfortable with the removal of an allowed set of frequencies for each variable. I'm not saying we shouldn't do this, but it does mean that any variable can be published at any frequency. It does give modelling groups more freedom to publish data, but it could lead to inconsistencies in the output of different groups. This is an item that I think warrants discussion, if only briefly, by the WIP.

I agree that users will be able to write a wider variety of unrequested datasets than in CIMP6.  In CMIP6, any variable found in any cmor table could be written even if that variable was not requested for a given experiment or a particular portion of an experiment.  And I'm sure lots of unrequested data was written.  Now a user will, in addition, be able to write data at unrequested frequencies.  If we are concerned that unrequested data (either from unrequested experiments or unrequested frequencies) will clutter up the archive, then we could write a bit of software whereby a user (or the publisher) could interrogate the data request and check whether a variable is in the data request (given an experiment, a time-slice, a frequency, and a region).  Then the that variable would be skipped.  I'm not sure who might step up and volunteer to do that,
but we could advertise, I suppose.

>     I'd be a little cautious about completing the max and min entries (fine for testing purposes) as there may (a) be a resolution dependence in certain variables and (b) be a frequency dependence (range of hourly data >> that of monthly means)

Definitely.  Given our current time-line, I agree that at most we should impose limits only on variables that are constrained physically by a max or min value (e.g., area fractions and other fractions, positive definite quantities like certain vertical fluxes and measures of quantity like mass which have a lower limit of 0, etc.).   Even then, some modelers will complain because in some models you can get things like slightly negative specific humidity (but values so small they don't affect anything).  If someone has the time and energy, they could also think of safe limits for ok_max_mean_abs and ok_min_mean_abs which would trap units problems in some variables.  For example, for land_area_fraction, if you set those limits at 100 and 2, you would trap someone trying to report this as a fraction and not a percentage, as requested.  (This test would raise a false error in a regional model place over the ocean (with no land) where the true mean land fraction is 0, independent of unit, which is less the ok_min_mean_abs values.

TestTables/CMIP7_CV.json
taylor13: yes, we'll likely do something like that for all the "simple" CVs, like this one.

TestTables/CMIP7_atmos2d.json
taylor13: that would be nice, but could be quite a big job, in FORTRAN at least.  Something, perhaps, later on.

taylor13:
Yes, @mauzey1, we need to add the nominal_resolution CV.  It has changed a bit from the [above](https://github.com/PCMDI/cmor/pull/778#issuecomment-2704636484).

durack1:
Additional tweaks

tweak
- [x] frequency (account for 3hr, 6hr pt values)

add
- [x] data_archive_id
- [x] nominal_resolution
- [x] region

TestTables/CMIP7_atmos2d.json
durack1: @mauzey1 it would be great if you can chime in on this query comment - how hard-coded are the data types in CMOR, and do we have the ability of using uint8 and other netCDF4 data types currently in CMOR?

TestTables/CMIP7_CV.json
```diff
@@ -0,0 +1,324 @@
...
+        },
+        "institution_id":{
+            "PCMDI":"Program for Climate Model Diagnosis and Intercomparison, Lawrence Livermore National Laboratory, Livermore, CA 94550, USA"
+        },
+        "license":[
+            "^CMIP7 model data produced by .* is licensed under a Creative Commons .* License (https://creativecommons\\.org/.*)\\. *Consult https://pcmdi\\.llnl\\.gov/CMIP7/TermsOfUse for terms of use governing CMIP7 output, including citation requirements and proper acknowledgment\\. *Further information about this data, including some limitations, can be found via the further_info_url (recorded as a global attribute in this file).*\\. *The data producers and data providers make no warranty, either express or implied, including, but not limited to, warranties of merchantability and fitness for a particular purpose\\. *All liabilities arising from the supply of the information (including any liability arising in negligence) are excluded to the fullest extent permitted by law\\.$"
+        ],
+        "mip_era":"CMIP7",
+        "nominal_resolution":[
+            "0.5 km",
```
durack1, line 176: add 0.25 km

TestTables/CMIP7_CV.json
```diff
@@ -0,0 +1,324 @@
...
+            "atmos":"Atmosphere",
+            "atmosChem":"Atmospheric Chemistry",
+            "land":"Land Surface",
+            "landIce":"Land Ice",
+            "ocean":"Ocean",
+            "ocnBgchem":"Ocean Biogeochemistry",
+            "seaIce":"Sea Ice"
+        },
+        "region":{
+            "antarctica":"located around the South Pole, separated from other land masses by the Southern Ocean, and almost entirely south of 60 degrees South latitude",
```
durack1, line 206: update glb, ant, gre, nhem, shem

TestTables/CMIP7_CV.json
```diff
@@ -0,0 +1,324 @@
...
+            "dec":{
+                "approx_interval":3650.0,
+                "description":"decadal mean samples"
+            },
+            "fx":"fixed (time invariant) field",
+            "mon":{
+                "approx_interval":30.0,
+                "description":"monthly mean samples"
+            },
+            "monC":{
```
durack1, line 148: lose monC

durack1:
> * I'm still a little uncomfortable with the removal of an allowed set of frequencies for each variable.  I'm not saying we shouldn't do this, but it does mean that any variable can be published at any frequency.  It does give modelling groups more freedom to publish data, but it could lead to inconsistencies in the output of different groups.  This is an item that I think warrants discussion, if only briefly, by the WIP.

Just replying to @matthew-mizielinski's comment above. The role of CMOR is to produce CF and project-compliant data (of which there are now quite a few projects other than CMIP), irrespective of whether this data has been requested by a project that is using the software. For e.g. data that was not included in the CMIP6 Data Request/DR was routinely produced for specific experiment(s), and made available either locally, or through the local node ESGF publication in some cases to those users.

The role of the ESGF publisher is to validate that any data that is being submitted to a project complies with project needs (global attributes, known variables, known frequencies, known activities/MIPs, known experiments required to populate the index matching search facets), and this could include a check whether certain data was requested by the CMIP7 project. While most CMIP contributors use CMOR all don't, which further suggests that the publisher (rather than CMOR) is the right place to do such project-compliance checks.

Ultimately, only data that is of broad interest (most likely requested in the CMIP7 DR) and produced by many groups will get replicated, and so I am less worried about your concern.

I've just finalized the "test" file formats, so will merge this PR, and we can continue this dialogue about compliance and how to enforce it elsewhere. This might be a good discussion to elevate to a WIP meeting agenda, rather squirreled away in a CMOR PR chatter

TestTables/CMIP7_atmos2d.json
durack1: Migrated this discussion into https://github.com/PCMDI/cmor/issues/781

PR #779, opened 2025-03-07
Handle variables that have a branding suffix

mauzey1:
These changes will allow CMOR to process variables that have a "branding suffix" in their table entry name as defined in #762.

There are also minor fixes to error messages for when`grid_label` or `nominal_resolution` are missing from the CV file.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -231,6 +231,7 @@ test_python: python
 	env TEST_NAME=Test/test_python_forecast_coordinates.py make test_a_python
 	env TEST_NAME=Test/test_cmor_CMIP6Plus.py make test_a_python
 	env TEST_NAME=Test/test_cmor_zstandard_and_quantize.py make test_a_python
+	env TEST_NAME=Test/test_python_branded_variable.py make test_a_python
 test_cmip6_cv: python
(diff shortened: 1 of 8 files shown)


durack1 (approved):
@mauzey1 great to see progress on this. I wonder how much does this depend on the format of CMIP7_CV.json in #778 - I am guessing it doesn't but just wanted to make sure

@taylor13 ping just FYI

mauzey1:
@durack1 This change will only get the branded variable from the variable tables. It doesn't check the attributes extracted from it against the CV. I can update this PR this week to include checking the CV.

taylor13:
regarding https://github.com/PCMDI/cmor/pull/779#issuecomment-2710784338, I think there are only a few values in the cmor tables that also appear in the CV lists of the CMIP7_CV.json file and I don't think CMOR needs to check those.  We can do those checks before putting a value of "approx_interval" or "realm" into a CMOR table.

We should, on the other hand check user input of such things as "source", "frequency", "region", etc. for consistency with the CVs. I think you're already doing that though.  Right?

mauzey1:
@taylor13 This PR will be merged to add the functionality for detecting the branding suffixes in variable names from the table and adding their values as global attributes. I will work on the CMIP7 checks in the next PR.

PR #780, opened 2025-03-09
configure: use correct CFLAGS and LDFLAGS variables

mkrupcale:
The `CFLAGS` and `LDFLAGS` variables were used opposite of their intended usage for the `udunits2` and `netcdf` external dependencies, so correct them.

 * `configure`: regenerate from...
 * `configure.ac`: use correct `CFLAGS` and `LDFLAGS` variables

### DIFF
--- a/configure
+++ b/configure
@@ -4891,12 +4891,12 @@ fi
      UDUNITS2LDFLAGS=" -L${with_udunits2}/lib -ludunits2"
    fi
   else
-    UDUNITS2FLAGS="-ludunits2"
-    UDUNITS2LDFLAGS=""
+    UDUNITS2LDFLAGS="-ludunits2"
+    UDUNITS2FLAGS=""
   fi
 else
-  UDUNITS2FLAGS="-ludunits2"
-  UDUNITS2LDFLAGS=""
+  UDUNITS2LDFLAGS="-ludunits2"
+  UDUNITS2FLAGS=""
 fi
 { printf "%s\n" "$as_me:${as_lineno-$LINENO}: checking for ut_parse in -ludunits2" >&5
 printf %s "checking for ut_parse in -ludunits2... " >&6; }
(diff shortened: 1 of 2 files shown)


durack1:
@mkrupcale thanks for this PR. Did you have some docs to point to which supports this flag correction?

It would also be really useful as background to understand where these settings caused issues, so that we could capture such a problem with a test moving forward

mkrupcale:
> @mkrupcale thanks for this PR. Did you have some docs to point to which supports this flag correction?

I think it's fairly standard that linker flags go in `LDFLAGS`, while C compiler flags go in `CFLAGS` [1], whereas for the `udunits2` and `netcdf` dependencies these were reversed. The other dependencies were using the correct ordering. Although, `LIBS` is probably the correct place for `-l` arguments in particular, but this PR is closer to correct than previously.

> It would also be really useful as background to understand where these settings caused issues, so that we could capture such a problem with a test moving forward

The reversal of compiler and linker flag variables could cause issues in the `Makefile`, although I didn't necessarily observe an issue directly because some linker flags can appear in compiler flag contexts and vice versa and may just be ignored in those contexts. I mainly just observed that the `configure` script output was reporting variables opposite of what I expected.

[1] https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.72/html_node/Preset-Output-Variables.html

durack1:
@mkrupcale did you test your changes in a build - looks like the change has broken our CI builds, see below

mauzey1:
@mkrupcale @durack1 I can confirm that this change works when building it on my machine.

I will approve and merge this change.

durack1:
> The CI is failing to checkout my `git` branch

@mauzey1 is there a trivial way to fix this? Can we cherry pick a patch from a remote repo for testing?

Issue #774, opened 2025-01-31
Adding/augmenting functions to write a CF Coordinate Reference System (CRS) "container variable"


durack1:
The [DRCDP project](https://github.com/PCMDI/DRCDP) has anticipated regional contributions (mainly for the CONUS) many of which in their native output include grid mapping criterion. While no obvious and extensive specification exists, some samples are available, including 3 defining a CRS WKT (CRS Well-Known Text format; also known as OGC WKT) within the [CF Conventions](https://cfconventions.org/cf-conventions/cf-conventions.html#use-of-the-crs-well-known-text-format) that define the specifications for the `"latitude_longitude"`, `"transverse_mercator"` using several common datums. The `int crs` variable written to a netCDF file has no associated data, but contains variable attributes which define characteristics of the grid mapping, .e.g., see [here](https://cfconventions.org/cf-conventions/cf-conventions.html#latitude-and-longitude-on-the-wgs-1984-datum-in-crs-wkt-format) and example below

```
  int crs ;
    crs:grid_mapping_name = "latitude_longitude";
    crs:longitude_of_prime_meridian = 0.0 ;
    crs:semi_major_axis = 6378137.0 ;
    crs:inverse_flattening = 298.257223563 ;
    crs:crs_wkt =
     GEODCRS["WGS 84",
     DATUM["World Geodetic System 1984",
       ELLIPSOID["WGS 84",6378137,298.257223563,
         LENGTHUNIT["metre",1.0]]],
     PRIMEM["Greenwich",0],
     CS[ellipsoidal,3],
       AXIS["(lat)",north,ANGLEUNIT["degree",0.0174532925199433]],
       AXIS["(lon)",east,ANGLEUNIT["degree",0.0174532925199433]],
       AXIS["ellipsoidal height (h)",up,LENGTHUNIT["metre",1.0]]]
```

In some parallel discussions see https://github.com/PCMDI/obs4MIPs-cmor-tables/issues/375, there may also be a need to account for "Equal-Area Scalable Earth" EASE grids (see [here](https://nsidc.org/data/user-resources/help-center/guide-ease-grids)). Ongoing discussions about the DRCDP targeted MACAv3 dataset (thanks @khegewisch) have noted the following datums:
1. WGS84 (EPGS: 4326) is the latitude_longitude that CMIP, and Google Earth uses
1. EPSG:3857 a Spherical Mercator projection coordinate system popularized by web services such as Google Maps, Leaflet Maps and OpenStreetMap
[...]

PR #786, opened 2025-03-13
Add CF grid mapping variants (#774)

durack1:
Ref #774 

@mauzey1 the CF grid mapping entries as found at http://cfconventions.org/Data/cf-conventions/cf-conventions-1.12/cf-conventions.html#appendix-grid-mappings. I hope this is a useful start


durack1:
Most of the information within this PR is already contained in the https://github.com/PCMDI/cmor/blob/3be56d2959734e0483bdc619d978cf2c788b6281/Src/cmor_grids.c#L382-L546 file - so will close without merging

PR #805, opened 2025-05-09
Only use zstandard compression if deflate is disabled and a valid compression level is provided

### LINKED ISSUE #798: Correct unexpected behavior of cmor.set_deflate(0,0,0)
The current CMOR version (3.10.0) has a quirk, that means when a user issues the command `cmor.set_deflate(varid, 0, 0, 0)` a file is generated which initiates zstandard compression, but a zero compression level (deflate level = 0, 4th arg). Upon opening these files with NCO, CDO, TempestExtremes, a `Filter error`is thrown, see below

The args for `cmor.set_deflate` are (var_id, shuffle, deflate, deflate_level) docs [here](https://cmor.llnl.gov/mydoc_cmor3_api/#cmor_set_deflate)

```python
This is what I see at the end:
 
                                :title = "STAR-ESDM 1.0 dataset prepared for DRCDP" ;
                                :tracking_id = "hdl:21.14100/959a1877-f48d-46e3-94a1-0d33c163f088" ;
                                :variable_id = "tasmax" ;
                                :license = "Creative Commons Attribution 4.0 International" ;
                                :cmor_version = "3.9.0" ;
data:
 
time_bnds =
NetCDF: Filter error: undefined filter encountered
Location: file vardata.c; fcn print_rows line 478
```

@mauzey1 has investigated the latest CMOR 3.10.0 code and concluded that "Looking into the code, the function for zstandard compression will get called whenever deflate is disabled.  So using deflate000 will cause the zstandard compression filters to be used with a compression level of 0.  To completely disable all compression, you will need to run cmor.set_deflate(0, 1, 0).  This will enable deflate, thus disabling the use of zstandard compression, but will not result in compression due to using a deflate level of 0."

More from @mauzey1:

Digging more into the code, CMOR will have deflate enabled by default with a default deflate level of 1.  If deflate is disabled by the user, then zstandard compression is applied with a default compression level of 3.  Below is the logic that determines when to use deflate or zstandard.
 
        // Only use zstandard compression if deflate is disabled
        if (deflate != 0) {
            ierr |= nc_def_var_deflate(ncid, var_id, shuffle, deflate, deflate_level);
        } else {
            ierr |= nc_def_var_deflate(ncid, var_id, shuffle, 0, 0);  // called to apply shuffle without deflate
            ierr |= nc_def_var_zstandard(ncid, var_id, zstandard_level);
        }
 
Maybe a better way of doing this is to have `nc_def_var_deflate` called first but then check if both deflate is disabled and zstandard’s level is within a valid range.
 
        ierr |= nc_def_var_deflate(ncid, var_id, shuffle, deflate, deflate_level);
        if (deflate != 0 && (zstandard_level >=-131072 && zstandard_level <= 22)) {
            ierr |= nc_def_var_zstandard(ncid, var_id, zstandard_level);
        }
 
The valid range of the zstandard compression level is [-131072, 22] according to [the doc for nc_def_var_zstandard](https://docs.unidata.ucar.edu/netcdf-c/4.9.2/dfilter_8c.html#a2841c7012840195954194f51f9750817).  If we set the default zstandard level to something out-of-range like -999999, then setting deflate to 0 would result in no compression being applied.  This would mean that a user must both disable deflate and set zstandard’s level to a valid value to enable zstandard compression.

@gleckler1 @paullric ping

mauzey1:
Resolves #798

CMOR by default uses deflation with a deflation level of 1.  To use zstandard compression, the user must disable deflation with `cmor_set_deflate` and then provide a valid zstandard compression level within the range [-131072, 22] using `cmor_set_zstandard`.  Performing only one of the two actions will result in no zstandard compression being applied.

Should there be a warning message for users who use out-of-range compression levels for `cmor_set_zstandard`?

### DIFF
--- a/Src/cmor.c
+++ b/Src/cmor.c
@@ -801,7 +801,7 @@ void cmor_reset_variable(int var_id)
     cmor_vars[var_id].shuffle = 0;
     cmor_vars[var_id].deflate = 1;
     cmor_vars[var_id].deflate_level = 1;
-    cmor_vars[var_id].zstandard_level = 3;
+    cmor_vars[var_id].zstandard_level = -99999;
     cmor_vars[var_id].quantize_mode = NC_NOQUANTIZE;
     cmor_vars[var_id].quantize_nsd = 1;
     cmor_vars[var_id].nomissing = 1;
@@ -2142,14 +2142,9 @@ int cmor_define_zfactors_vars(int var_id, int ncid, int *nc_dim,
                         icz =
                           cmor_tables[nTableID].vars[nTableID].zstandard_level;
 
-                        if (icd != 0) {
-                            ierr |= nc_def_var_deflate(ncid, nc_zfactors[lnzfactors],
-                                                    ics, icd, icdl);
-                        } else {
-                            ierr |= nc_def_var_deflate(ncid, nc_zfactors[lnzfactors],
-                                                    ics, 0, 0);
-                            ierr |= nc_def_var_zstandard(ncid, nc_zfactors[lnzfactors],
-                                                    icz);
+                        ierr |= nc_def_var_deflate(ncid, nc_zfactors[lnzfactors], ics, icd, icdl);
+                        if (icd == 0 && (icz >= -131072 && icz <= 22)) {
+                            ierr |= nc_def_var_zstandard(ncid, nc_zfactors[lnzfactors], icz);
                         }
 
                         if (ierr != NC_NOERR) {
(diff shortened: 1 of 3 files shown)


durack1 (changes requested):
@mauzey1 exactly what I was thinking, having an error (or warning, if it's not going to work as expected I suggest we error and exit) thrown when a user chooses and out-of-range compression level would be exactly what I would recommend we add

mauzey1:
> @mauzey1 exactly what I was thinking, having an error (or warning, if it's not going to work as expected I suggest we error and exit) thrown when a user chooses and out-of-range compression level would be exactly what I would recommend we add

`cmor_set_zstandaard` will now throw the following for out-of-range levels.
```
C Traceback:
In function: cmor_set_zstandard
! 

!!!!!!!!!!!!!!!!!!!!!!!!!
!
! Warning: Your zstandard level of 100 is outside of the range [-131072, 22].
! Zstandard compression will not be applied to variable id(0)
!
!!!!!!!!!!!!!!!!!!!!!!!!!
```

mauzey1:
@durack1 Are you fine with the warning that I added to `cmor_set_zstandard`?  Should it be elevated to an error?

durack1:
> @durack1 Are you fine with the warning that I added to `cmor_set_zstandard`? Should it be elevated to an error?

I agree with you, as this will generate results that are outside of the range of the function, this should be an error, not a warning - we want to catch any cases of bad data BEFORE they are written - thanks!

mauzey1:
Should it be an error that halts the program (i.e. CMOR_CRITICAL)?

durack1:
> Should it be an error that halts the program (i.e. CMOR_CRITICAL)?

Yes, I think so

mauzey1:
@durack1 The warning has been changed to an error.  There is also a test that checks that this error is thrown.

durack1 (approved):
Wonderful, thanks @mauzey1

PR #815, opened 2025-05-24
Catch unsupported CF calendars as errors

### LINKED ISSUE #813: Update calendar support to match full CF 1.12 list
> @mauzey1 I was surprised to see that `none` is in fact a valid CF calendar - see [here](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.12/cf-conventions.html#none-calendar). Is the monotonic value error a code problem do you think?
> 
> I've just consulted the latest CF 1.12, and noted a couple of additional calendars have now appeared. The full list includes: `standard`, `proleptic_gregorian`, `julian`, `utc`, `tai`, `noleap/365_day`, `all_leap/366_day`, `360_day`, `none` - documented [here](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.12/cf-conventions.html#calendar). Do we have full calendar coverage in the current CMOR code?
> 
> In addition, do we have any other instances of old CDMS behaviours lurking around?
> 
> If there are any changes required, should we reopen this issue and tag against a CMOR 3.11? 

 _Originally posted by @durack1 in [#800](https://github.com/PCMDI/cmor/issues/800#issuecomment-2902254110)_

The amendment to CMOR is as below

[@durack1](https://github.com/durack1) Here is a list of supported calendars in CMOR. It does not contain utc, tai, or all_leap/366_day. Do we want them added?

https://github.com/PCMDI/cmor/blob/6acabca7fc5c2c06c1c39feb9f4ae36e19a73103/Src/cmor.c#L82-L91

I think it is likely that anyone specifying these odd calendars for MIP experiments or obs4MIPs is doing so by mistake, so we should raise an error and say: "an XXX calendar is not currently an option for MIP results. Please adopt a different, more common calendar used for climate studies".
where XXX is utc, tai, or all_leap/366_day.

Yes exactly. CMOR should recognize all CF-valid calendars, but only allow the use of calendars that make sense within the MIP context, and so raise an error as you say.

[@taylor13](https://github.com/taylor13) is there a valid use of the none calendar that we have, thinking obs4MIPs might be the most likely climatology use?

mauzey1:
Resolves #813 

If a CF calendar name that is unsupported by CMOR is used for the attribute `calendar`, then CMOR will throw an error.

```
C Traceback:
! In function: cmor_calendar_c2i
! called from: cmor_close_variable
! 

!!!!!!!!!!!!!!!!!!!!!!!!!
!
! Error: A utc calendar is not currently an option for MIP results.
! Please adopt a different, more common calendar used for climate studies
!
!!!!!!!!!!!!!!!!!!!!!!!!!
```

### DIFF
--- a/Src/cmor_axes.c
+++ b/Src/cmor_axes.c
@@ -8,9 +8,11 @@
 
 int cuErrOpts = CU_VERBOSE;
 
-/************************************************************************/
-/*                         cmor_calendar_c2i()                          */
-/************************************************************************/
+/*****************************************************************************/
+/*                         cmor_calendar_c2i()                               */
+/* This function recognizes all valid calendars in the CF 1.12 release       */
+/* See https://cfconventions.org/cf-conventions/cf-conventions.html#calendar */
+/*****************************************************************************/
 int cmor_calendar_c2i(char *calendar, cdCalenType * ical)
 {
 
@@ -39,10 +41,13 @@ int cmor_calendar_c2i(char *calendar, cdCalenType * ical)
             || strcmp(calendar, "all_leap") == 0
             || strcmp(calendar, "366_day") == 0) {
         cmor_handle_error_variadic(
-            "A %s calendar is not currently an option for MIP results."
+            "A %s calendar is an invalid selection for Model Intercomparison Project (MIP) data."
             "\n! "
-            "Please adopt a different, more common calendar used for "
-            "climate studies", CMOR_NORMAL, calendar);
+            "Please adopt a different, more common calendar used for climate studies. "
+            "\n!\n! "
+            "This function recognizes all valid calendars in the CF 1.12 release. "
+            "See https://cfconventions.org/cf-conventions/cf-conventions.html#calendar",
+            CMOR_NORMAL, calendar);
         cmor_pop_traceback();
         return (1);
     } else {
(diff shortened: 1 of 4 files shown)


taylor13:
this looks good to me.

Src/cmor_axes.c
```diff
@@ -34,7 +34,18 @@ int cmor_calendar_c2i(char *calendar, cdCalenType * ical)
         *ical = cdJulian;
     else if (strcmp(calendar, "none") == 0)
         *ical = cdClim;
-    else {
+    else if (strcmp(calendar, "utc") == 0
+            || strcmp(calendar, "tai") == 0
+            || strcmp(calendar, "all_leap") == 0
+            || strcmp(calendar, "366_day") == 0) {
+        cmor_handle_error_variadic(
+            "A %s calendar is not currently an option for MIP results."
```
durack1, line 42: How about "A %s calendar is an invalid selection for Model Intercomparison Project (MIP) data."

durack1 (approved):
@mauzey1 wonderful - looks great, a little tweak and we're good to merge

Src/cmor_axes.c
durack1: Can we also augment the function info with:

```python
/************************************************************************/
/*                         cmor_calendar_c2i()                          */
/* This function recognizes all valid calendars in the CF 1.12 release */
/************************************************************************/

PR #823, opened 2025-06-10
Only add license attributes to a file if they are required by the CV

### LINKED ISSUE #811: Update license_id entry - CMOR 3.10 changes break test
It seems our test case https://github.com/PCMDI/cmor/blob/main/Test/test_cmor_CMIP7.py is currently broken - due to the CMOR 3.10 license_id changes - can we get that tweaked and working ASAP please @mauzey1

mauzey1:
Resolves #811 

If the user or the CV define the `license_id`, `license_type`, or `license_url` attributes used for creating a license string, they will only be added as global attributes in a file if they are listed in the `required_global_attributes` section of a CV.


durack1 (approved):
@mauzey1 great will also pull this down once it's available. My python demos will work fine with this

PR #827, opened 2025-06-16
Check user-defined required attributes before checking all required attributes

### LINKED ISSUE #737: order in `required_global_attributes` matters
Just wanted to dump this here, since we now have fixed our [required global attributes for CORDEX](https://github.com/WCRP-CORDEX/cordex-cmip6-cv/blob/main/CORDEX-CMIP6_required_global_attributes.json). I'm not sure if it's by intention, but it seems like the order of variables in the `required_global_attributes` tables seems to matter when they are evaluated during cmorization. This is probably related to https://github.com/PCMDI/cmor/issues/628. For example, if if have...

```json
{
    "CV": {
        "required_global_attributes": [
            "domain_id",
            "domain",
        ],
        "domain_id": {
            "EUR-50": {
                "domain": "Europe",
                "domain_id": "EUR-50"
            }
        }
    }
}
```
... and my input dataset file contains `domain_id: "EUR-50"`, cmor will correctly and automatically add the global attributes `domain: "Europe"` to the file. However, this wil not work if the `domain` attribute comes before the `domain_id` attribute in the list of `required_global_attributes`, e.g., this will not work...

```json
"required_global_attributes": [
       "domain",
       "domain_id",
],
```
This is somehow confusing and it's not the case for attributes like, e.g., `source_id` and `source`....

mauzey1:
Resolves #737

CMOR will now check the required attributes set by the user first before checking all required attributes listed in `required_global_attributes` in the CV. This allows for the setting of attributes nested within attributes such as the `domain_id` example in https://github.com/PCMDI/cmor/issues/737#issue-2240078052. In that example, `domain_id` will be validated first to allow the setting of the `domain` attribute. A second validation pass will then validate the `domain` attribute.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -238,6 +238,7 @@ test_python: python
 	env TEST_NAME=Test/test_cmor_unsupported_calendar.py make test_a_python
 	env TEST_NAME=Test/test_cmor_time_interval_check.py make test_a_python
 	env TEST_NAME=Test/test_cmor_license_attributes.py make test_a_python
+	env TEST_NAME=Test/test_cmor_nested_cv_attribute.py make test_a_python
 test_cmip6_cv: python
 	env TEST_NAME=Test/test_python_CMIP6_CV_sub_experimentnotset.py make test_a_python
 	env TEST_NAME=Test/test_python_CMIP6_CV_sub_experimentbad.py make test_a_python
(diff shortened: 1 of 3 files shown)


durack1:
@larsbuntemeyer it would be useful to get your chime-in here, this change intends to resolve your issue defined in #737

PR #838, opened 2025-07-08
Warn user when the CV contains attribute values that are invalid for CMOR 

### LINKED ISSUE #829: Catching issues with poorly formed CV data
We've seen a couple of cases of poorly formed input CV files not causing CMOR to fail, but confusing users. In these cases, CMOR runs without reporting a warning or error, BUT, it doesn't do what a user would expect, which is validate entries correctly.

A good example of this, is captured in the obs4MIPs repo (see https://github.com/PCMDI/obs4MIPs-cmor-tables/issues/485#issuecomment-2977424403). In summary, this was due to an additional nesting in the `obs4MIPs_CVs.json` file, which mean that rather than the nominal_resolution containing a list of valid values to check against, a dictionary was returned, and CMOR didn't interrogate any information at all.

In brief, the issue is the double-nesting in the CVs file, see below (and [here](https://github.com/PCMDI/obs4MIPs-cmor-tables/blob/fb7fb3f4dcce0106434680931b71518b27e81955/Tables/obs4MIPs_CV.json#L127-L144)):
```python
"nominal_resolution":{
    "nominal_resolution":[
        "0.5 km",
        "1 km",
        "10 km",
...
```

This should have been:
```python
"nominal_resolution":[
    "0.5 km",
    "1 km",
    "10 km",
...
```

Admittedly, better checking of the input data preparation could have solved this problem, but it's not the first time I have seen this. Consequently it would be great if for this particular case (and for cases like it), a warning would be thrown, indicating that (in this case) the nominal_resolution CV provided is an unexpected data structure (a JSON dictionary) and no valid values will be checked as a consequence. Such a warning would hopefully trigger off a user to an input CV (or an input file) error, this can be identified, resolved and happy CMORizing will be the result.

@mauzey1 let me know if the above needs further expansion or if other cases should be added.

In CMOR 3.11 for CMIPx related CVs, 3 are expected as complex/nested structures, the experiment_id, the source_id and the license_id. All others are expected to be simple non complex dictionary (key:value pairs) or in the case of the `nominal_resolution` above a list.

@taylor13 @gleckler1 @matthew-mizielinski @jamesAnstey if you folks have any other examples you'd like to provide for a test suite, we're all ears

mauzey1:
Resolves #829 

CMOR will throw a warning if attribute values in a CV are not in their expected format.  For example, `mip_era` is expected to be a single string value rather than an array or object.  CMOR will also throw a warning for JSON structures that are not valid for the CV such as arrays that have other arrays or objects as elements.


mauzey1:
@durack1 I'm going to set this PR to a draft since there are some refinements I want to apply to the check, such as checking that values in a list are strings since most of the attribute checks are string based.

durack1:
> I'm going to set this PR to a draft since there are some refinements I want to apply to the check, such as checking that values in a list are strings since most of the attribute checks are string based.

Sounds perfect to me - happy to review when you're ready. I think being as careful as we can be is a good move here - if we have a `nightly` to test that would be ideal too

mauzey1:
@durack1 I've updated this PR so that the CV check is more inclusive of current CV formats and more reflects how CMOR processes them.  I've also added a check for the `CV` section in a CV file.

PR #848, opened 2025-07-18
Don't show warning if standard_parallel1 and standard_parallel2 are used for grid mapping/CRS

### LINKED ISSUE #845: Erroneous missing `standard_parallel` warning raised when two standard parallel values are passed
For both `cmor_set_grid_mapping` and `cmor_set_crs`, a warning is raised for the attribute `standard_parallel` being missing even if `standard_parallel1` and `standard_parallel2` were used for a `standard_parallel` that has two values.

```
C Traceback:
In function: cmor_set_grid_mapping
! called from: cmor_set_grid_mapping
! 

!!!!!!!!!!!!!!!!!!!!!!!!!
!
! Warning: Grid mapping attribute standard_parallel has not been set, you should consider setting it
!
!!!!!!!!!!!!!!!!!!!!!!!!!
```

The warning for missing the `standard_parallel` attribute should be something like below.
```
Warning: Grid mapping attribute standard_parallel has not been set, you should consider setting standard_parallel for one value or
standard_parallel1 and standard_parallel2 for two values
```

@durack1 The below section of code shows that only the grid mapping `lambert_conformal_conic` can use two standard parallel values.  Is this still true?  According to the CF conventions, grid mapping [`albers_conical_equal_area`](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.12/cf-conventions.html#_albers_equal_area) can also have a `standard_parallel` with 1 or 2 values.
https://github.com/PCMDI/cmor/blob/def02f29c7311bbc64368ffbe4f22555a316ea70/Src/cmor_grids.c#L731-L750

mauzey1:
Resolves #845 

A warning message won't be displayed by `cmor_set_grid_mapping` or `cmor_set_crs`  for a missing `standard_parallel` parameter if passing two values with `standard_parallel1` and `standard_parallel2`.

This PR also allows the `albers_conical_equal_area` grid mapping to have two standard parallel values.

### DIFF
--- a/Src/cmor_grids.c
+++ b/Src/cmor_grids.c
@@ -388,6 +388,12 @@ int cmor_grid_valid_mapping_attribute_names(char *name, int *natt, char (*att)
                                             char (*dims)
                                             [CMOR_MAX_STRING])
 {
+/* -------------------------------------------------------------------- */
+/*   The grid mapping attribute mappings below are consistent with      */
+/*   CF 1.12 (https://cfconventions.org/Data/cf-conventions/            */
+/*   cf-conventions-1.12/cf-conventions.html#appendix-grid-mappings)    */
+/* -------------------------------------------------------------------- */
+
     int i, j;
 
     *natt = -1;                 /* -1 means mapping name not found */
--- a/Src/cmor_grids.c
+++ b/Src/cmor_grids.c
@@ -809,7 +809,7 @@ int cmor_set_crs(int gid, char *grid_mapping, int nparam,
     char *achar, *bchar;
     char ltext_attributes_name[CMOR_MAX_STRING];
 
-    cmor_add_traceback("cmor_set_grid_mapping");
+    cmor_add_traceback("cmor_set_crs");
 
     ierr = cmor_set_grid_mapping(gid, grid_mapping, nparam,
         attributes_names, lparams, attributes_values,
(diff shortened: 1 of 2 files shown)


Src/cmor_grids.c
```diff
@@ -760,11 +762,24 @@ int cmor_set_grid_mapping(int gid, char *name, int nparam,
 
     for (i = 0; i < nattributes - 7; i++) {
         if (cmor_has_grid_attribute(gid, grid_attributes[i]) == 1) {
-            cmor_handle_error_variadic(
-                "Grid mapping attribute %s has not been set, "
-                "you should consider setting it",
-                CMOR_WARNING,
-                grid_attributes[i]);
+            if (strcmp(grid_attributes[i], "standard_parallel") == 0) {
```
durack1, line 765: @mauzey1 would it be useful to note within the code that the settings for the included grid_mappings are consistent with CF 1.12? These will likely to continue to evolve, so we'd want to attempt to keep track of version changes, and note our current baseline

durack1 (approved):
@mauzey1 just a comment to add a reference to current CF 1.12, which is what we've mapped the argument requirements from

Src/cmor_grids.c
mauzey1: Do you want something to the effect of the following in `cmor_grid_valid_mapping_attribute_names`?
```
/* -------------------------------------------------------------------- */
/*   The grid mapping attribute mappings below are consistent with      */
/*   CF 1.12 (https://cfconventions.org/Data/cf-conventions/            */
/*   cf-conventions-1.12/cf-conventions.html#appendix-grid-mappings)    */
/* -------------------------------------------------------------------- */

```

durack1: Wonderful, thanks

PR #855, opened 2025-08-11
Check parameter options for `mercator` and `polar stereographic` grid mappings

### LINKED ISSUE #849: Parameter options for `mercator` and `polar_stereographic` grid mappings
According to CF 1.12, [`mercator`](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.12/cf-conventions.html#_mercator) and [`polar_stereographic`](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.12/cf-conventions.html#polar-stereographic) both have the option to use either `standard_parallel` or `scale_factor_at_projection_origin` as a parameter.

Currently, CMOR requires `mercator` to use both `standard_parallel` or `scale_factor_at_projection_origin` while `polar_stereographic` only allows `standard_parallel`.

Users should be allowed to set either `standard_parallel` or `scale_factor_at_projection_origin` for these grid mappings and not require both.  Should we allow both?

mauzey1:
Resolves #849 

Warns users if neither `standard_parallel` or `scale_factor_at_projection_origin` were set as parameters for grid mappings `mercator` and `polar stereographic`.  It also warns users if both parameters were set on the same grid mapping.

### DIFF
--- a/Test/test_cmor_crs.py
+++ b/Test/test_cmor_crs.py
@@ -3,7 +3,8 @@ import unittest
 import numpy as np
 from netCDF4 import Dataset
 from base_CMIP6_CV import BaseCVsTest
-from test_python_common import *
+from test_python_common import gen_irreg_grid, read_2d_input_files, \
+                               Time, bnds_time, varin2d, lat, lon
 
 
 class TestCRS(BaseCVsTest):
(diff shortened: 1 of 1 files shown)


durack1 (approved):
@mauzey1 wonderful!

PR #863, opened 2025-08-29
Check if nominal_resolution value is found in a dictionary in the CV in grid check

mauzey1:
`nominal_resolution` in the CV should be either an array or dictionary with key-value pairs.  However, the grid checking function only looks for an array. This quick patch adds the check for the dictionary.

It's a bit redundant since the issue will also get caught by the global attribute check but I wanted the check to match the one done for `grid_label` in the same function.

Not a blocker for the CMOR 3.12 release.


durack1 (approved):
@mauzey1 nice catch

PR #867, opened 2025-10-01
Clean up doc strings in pywrapper.py

### LINKED ISSUE #866: Usage incorrect for python set_variable_attribute
Documentation shows:     Usage:                                                                                                                                                      
      cmor.set_variable_attribute(var_id,name,value)     

However the interface is:
def set_variable_attribute(var_id, name, data_type, value):

https://github.com/PCMDI/cmor/blob/main/Lib/pywrapper.py#L1061

mauzey1:
Resolves #866

### DIFF
--- a/Lib/pywrapper.py
+++ b/Lib/pywrapper.py
@@ -74,8 +74,8 @@ def grid(axis_ids, latitude=None, longitude=None,
     longitude_vertices=None)
     Where:
     axis_ids : array contianing the axes ids for this grid.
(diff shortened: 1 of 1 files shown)


taylor13 (approved):
These all look like simple corrections.  I won't vouch for the syntax, but they all appear to be correct.  If you are uncertain as to their correctness, I would ask Matt to look them over, but I think you can go ahead and "commit" or "merge" or whatever the proper term.

Issue #861, opened 2025-08-28
Handling of secondary realm for CMIP7 output


matthew-mizielinski:
The following just came up for discussion in the DR TISG

The CMIP7 style tables are organised by realm and have a single realm entry available for each table.

Many variables in CMIP6 and CMIP7 have secondary realms as they relate to the interface between model components. An example is the CMIP6 variable `Omon/sfdsi` (Downward Sea Ice Basal Salt Flux), which had realm "ocean seaice".  With the current system the secondary "seaice" realm reference is discarded and in the CMIP7 prototype tables this would appear in the CMIP7_ocean table via the variable `sfdsi_tavg-u-hxy-sea`.

I don't think with the CMIP7 style tables there is a means to include secondary realms.

Many members of the TISG commented that it would be valuable to retain the secondary realms in the output files -- is this something that could be added to the CMOR library when working with CMIP7?

Given that the realm is now used to organise the variables into tables is it possible to include a secondary realm field for variables that could be populated if additional realms are needed? 

I don't think that extending the global attribute `realm` is problematic as `table_id` is being used as part of the DRS (I think).

@durack1 @jamesanstey @taylor13  -- thoughts welcome, but it might be worth resolving this before v3.12 is released.

mauzey1:
@matthew-mizielinski I was rearranging the milestones thinking that we can get changes for https://github.com/PCMDI/cmor/issues/864 into a release sooner.  I created the 3.14 milestone to put the other changes in the next release.  I will start looking at this issue now and place it in the 3.13 release.  We'll aim for a release date of next Thursday, October 9.

mauzey1:
Looking at our current tables, we do have `modeling_realm` for variable `thetao_tavg-l-hxy-sea` in CMIP7_oceanLev.json.
https://github.com/PCMDI/cmor/blob/47d7e333a8357684c1bbb40d63a1d440dfe1a359/TestTables/CMIP7_oceanLev.json#L32-L34
`modeling_realm` has been used by CMIP6 variables to define multiple realms in a space-separated string, which is then stored in the global attributes of netCDF files.  In the case of variable tables that store multiple `modeling_realm` values in arrays, they are turned into space-separated strings by CMOR to be stored in a similar manner as CMIP6 realms.

The variable `pr_tavg-u-hxy-u` in CMIP7_atmos2d.json has a `realm` attribute that contains multiple values.
[...]

PR #871, opened 2025-10-06
Fix parsing of branding suffixes and add tests for secondary realms CMIP7

### LINKED ISSUE #870: Processing multiple variable tables with branding suffixes corrupts labels and out names
String values for branding suffix, labels, and out name are being stored in the variable definition without terminating null characters.  This causes strings for these attributes to terminate at the previously-set null terminator. For example, if the last `area_label` was `tree` and the next `area_label` is `u` then you might end up with a value `uree`.

The processing of the branding suffix occurs in this section:
https://github.com/PCMDI/cmor/blob/47d7e333a8357684c1bbb40d63a1d440dfe1a359/Src/cmor_tables.c#L218-L243

Two potential solutions:
1. Add null characters to the end of each string.
2. Initialize all strings in the variable definition to have all null characters.

The former should be easy, but the latter might help with other parts of CMOR where similar issues might occur.  Maybe we could just apply both for extra safety.

mauzey1:
Fixes #870 

Related to #861

@matthew-mizielinski


mauzey1:
I've updated this to include having the string values in the variable definition be initialized with all null characters to flush out any residual values.

PR #894, opened 2026-01-05
New chunking strategy to support CMIP7 repacking requirements

### LINKED ISSUE #880: Repacking
Hi,

I just learned "repacking" should be integrated into CMOR which would be great in a sense, since, imo, noone would understand why we need two tools for data standardization and publication.

Is there a plan when that can be realized? We may wait until then with CMIP7 publication.

Best,
Fabi

mauzey1:
Resolves #880 

This changes how CMOR performs chunking of the variable data along time and spatial dimensions.

Key changes

- Chunking dimensions of the time and time bounds axes will be set to the length of the time axis, plus a dimension of size 2 for the bounds, if it is known when the time axis is created.  Otherwise, the time and time bounds chunking dimensions will be set to 512 by default.
- Chunking dimensions for data variables will be determined by whether chunks can be at least 4 MB in size.
  - If multiple time slices of data can fit within 4 MB, then chunking dimensions will have multiple time slices and full spatial dimensions of the data variable.
  - If a single time slice is greater than 4 MB, then chunking dimensions will have one time slice with spatial chunking dimensions determined by the netcdf library.

These changes do not enforce the "consolidated metadata" repacking requirement where the chunking index b-tree must be stored contiguously at the beginning of the file.  However, it should reduce the work needed to be done by [cmip7repack](https://github.com/NCAS-CMS/cmip7_repack) by avoiding rechunking the data variable.


durack1 (approved):
@mauzey1 looks great!

PR #896, opened 2026-01-08
Allow users to set the value of variable attribute `long_name`

### LINKED ISSUE #891: Allow users of CMOR to override long_name from tables
There has been some discussion on the WIP about overriding the the use of long names when writing data for CMIP7. At present this does not appear to be possible with CMOR as the `variable()` routine does not allow this field to be specified via a keyword argument and `set_variable_attribute()` is not allowed to override it.

Would it be acceptable to allow the user to override `long_name` variable attributes when writing data?

If I'm reading the code correctly then this would simply require the removal of the following line from the `set_variable_attribute` routine;

https://github.com/PCMDI/cmor/blob/cb5066bac3d9d82eee3bf2946aa54cea5ef2b20a/Src/cmor_variables.c#L208

@taylor13 @durack1 @jamesanstey thoughts welcome

mauzey1:
Resolves #891

### DIFF
--- a/Src/cmor_variables.c
+++ b/Src/cmor_variables.c
@@ -205,7 +205,6 @@ int cmor_set_variable_attribute(int id, char *attribute_name, char type,
         (strcmp(attribute_name, VARIABLE_ATT_MISSINGVALUES) == 0) ||
         (strcmp(attribute_name, VARIABLE_ATT_FILLVAL) == 0) ||
         (strcmp(attribute_name, VARIABLE_ATT_STANDARDNAME) == 0) ||
-        (strcmp(attribute_name, VARIABLE_ATT_LONGNAME) == 0) ||
         (strcmp(attribute_name, VARIABLE_ATT_FLAGVALUES) == 0) ||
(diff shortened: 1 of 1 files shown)


Src/cmor_variables.c
```diff
@@ -205,7 +205,6 @@ int cmor_set_variable_attribute(int id, char *attribute_name, char type,
         (strcmp(attribute_name, VARIABLE_ATT_MISSINGVALUES) == 0) ||
         (strcmp(attribute_name, VARIABLE_ATT_FILLVAL) == 0) ||
         (strcmp(attribute_name, VARIABLE_ATT_STANDARDNAME) == 0) ||
-        (strcmp(attribute_name, VARIABLE_ATT_LONGNAME) == 0) ||
```
durack1: @mauzey1 wow, is this really the only change required to enable a user to overwrite this attribute? If yes, it makes me a little queasy that we also allow standard_name to also be overwritten (if my wild assumptions about entries in this table enabling users to overwrite the table-provided values is correct

PR #906, opened 2026-01-26
Add `cmor_set_chunking` function

### LINKED ISSUE #895: Allow users to manually define variable chunking through the API
> From the discussion above, with good points on the value of data chunks containing multiple times, I'm starting to think that this is something to give the user of CMOR control over.
> 
> Would it be possible to add in a function call that allows us to set (a) the number of time records per chunk and (b) the total chunk size?  
> 
> e.g. `cmor_set_chunking(var_id, time_chunk_size, max_num_chunk_points)` or `cmor_set_chunking(var_id, chunk_shape_tuple)`
> 
> The user would then be able to determine the chunking for the variable being written and optimise appropriately. In the longer term it would be valuable to provide guidance to groups on the optimum strategy.
>  

 _Originally posted by @matthew-mizielinski in [#880](https://github.com/PCMDI/cmor/issues/880#issuecomment-3612288495)_

mauzey1:
Resolves #895 

This adds the function `cmor_set_chunking` to the API, allowing users to set the NetCDF chunking dimensions of variables manually.

This PR also adds a minor tweak to the Fortran API. You can now specify the use of NetCDF 4 functionality in Fortran programs with `cmor_setup` by setting the netcdf_file_action parameter to "preserve_4", "append_4", or "replace_4".

### DIFF
--- a/Lib/pywrapper.py
+++ b/Lib/pywrapper.py
@@ -1123,7 +1123,7 @@ def set_chunking(var_id, chunking_dimensions):
         positive integers that is the same length as the number of axes
         for the variable. The list must follow the order of the
         variable's list of axes.
-        
+
         Any negative values in chunking_dimensions will make CMOR warn
         that the value will be replaced with zero.
 


durack1 (approved):
@mauzey1 looks good, thanks!

PR #907, opened 2026-01-28
Resolve 'Error using `store_with` option in cmor_write (cmor 3.13)'

### LINKED ISSUE #902: Error using `store_with` option in cmor_write (cmor 3.13)
Hi CMOR folks,

I'm working on a CMORization pipeline for CCCma's CanESM model for CMIP7, and am encountering an error when using the `store_with` argument to `cmor.write()` to store the surface pressure alongside atmosphere variables on hybrid levels. The text of the error is, for example:
```
! Error: while closing variable 4 (so2, table aerosol) we
! noticed it has a time varying associated
! variable, you wrote 12 time steps for the
! variable, but its associated variable 5 (ps)
! has 0 times written
```
and comes from the call to `cmor.close`.

I am a bit perplexed by the error message. When this occurs, the output file is created, and the surface pressure `ps` does in fact have data for all time steps. The only issue seems to be that the date string in the file name is mangled, e.g., `so2_tavg-al-hxy-u_mon_glb_gn_CanESM5-1_piControl_r1i1p1f19mvuNY3571014.nc`--the `9mvuNY3571014` should be a date string like `_185001-185012`. 

I did spot the error in a couple of past issues in this repo: #150 and #814, though the context in those is a bit different 

The scripts that I'm using for this are rather old (and not written by me) and were previously used for CanESM for CMIP6 with CMOR v3.9. I am in the process of updating these for v3.13. I can recreate the steps taken back then, with CMOR v3.9, and there the error does not occur. 

So, I thought I'd test the [example provided on the CMOR docs](https://cmor.llnl.gov/mydoc_cmor3_python/#example-6-treatment-of-a-3-d-field-on-model-levels-cloud-fraction-a-function-of-longitude-latitude-model-level-month) to see if that still works. Of course, that example may be out-of-date, but the error appears there too. Again, it occurs in CMOR v3.13, but not in v3.9. 

Between v3.9 and v3.13, has something changed about the syntax or behavior of cmor.zfactor or cmor.write that could cause this? 

Thanks,
Russell Deitrick

mauzey1:
Fixes #902 

The associated times written count for the reference variable will be updated with the number of times written for the associated variable whether or not times are passed to `cmor_write`.

This also adds [example6.py](https://github.com/PCMDI/cmor3_documentation/blob/0a31b73d9787419c1df97ee9d16d8b0a87d651eb/mydoc/examples/example6.py) from the doc site as a unit test.

### DIFF
--- a/Lib/__init__.py
+++ b/Lib/__init__.py
@@ -13,7 +13,7 @@ from cmor.pywrapper import (
     has_cur_dataset_attribute, set_variable_attribute, get_variable_attribute,
     has_variable_attribute, get_final_filename, set_deflate, set_zstandard,
     set_quantize, set_furtherinfourl, set_climatology, get_climatology,
-    set_terminate_signal, get_terminate_signal)
+    set_terminate_signal, get_terminate_signal, set_chunking)
 
 try:
     from check_CMOR_compliant import checkCMOR
--- a/Lib/pywrapper.py
+++ b/Lib/pywrapper.py
@@ -1112,6 +1112,26 @@ def set_quantize(var_id, quantize_mode, quantize_nsd):
     return _cmor.set_quantize(var_id, quantize_mode, quantize_nsd)
 
 
+def set_chunking(var_id, chunking_dimensions):
+    """Sets chunking dimensions on a cmor variable
+    Usage:
+      cmor.set_chunking(var_id, chunking_dimensions)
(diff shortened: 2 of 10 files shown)


durack1 (approved):
@mauzey1 nice to clean up these loose ends!

PR #908, opened 2026-01-29
Remove Python 2 C API code from wrapper code

mauzey1:
Finish removing Python 2 supporting code that was left over after the transition from Python 2 to 3.


durack1 (approved):
@mauzey1 great, thanks for catching this - great to keep cleaning and improving

PR #940, opened 2026-04-14
Only write parent attributes when they are required for parent experiments in CMIP7 datasets

### LINKED ISSUE #937: urgent release requested
@mauzey1  @matthew-mizielinski: 

Following a recent review of the output produced by CMOR, we have discovered some issues that we need to quickly address. CMOR is currently producing output files that are not quite compliant with the CMIP7 requirements.  There are some additional changes that I think we should make but perhaps are not essential.  Below all the recognized changes are listed roughly in order of their priority:

1. REQUIRED NOW:  In CMIP7, the global attributes documenting the "parent" experiment and branch time should only be recorded in files when the parent exists;  unlike CMIP6 where these were set to "no parent" (in most cases), in CMIP7 the attributes should not exist in the file.  I think CMOR current knows whether or not an experiment has a parent, so perhaps it won't be too difficult for it to simply skip writing the following global attributes if the parent experiment should not exist:
- branch_time_in_child
- branch_time_in_parent
- parent_activity_id
- parent_experiment_id
- parent_mip_era
- parent_source_id
- parent_time_units
- parent_variant_label
There may be alternative ways to achieve the same ends, but once we do this, the CMOR-produced files will comply with the CMIP7 specifications.
2.  DETERMINE NOW:  How does CMOR know the directory path to the directory where it will write files?  As I recall the user can somehow override where CMOR would put output in the cmor.setup function, but I can't remember how this works.  In particular, if CMOR doesn't receive any guidance from the user, how does it create the "version" subdirectory in the template: "<drs_specs><mip_era><activity_id><institution_id><source_id><experiment_id><variant_label><region><frequency><variable_id><branding_suffix><grid_label><version>"?  The user must be able to override "version" if "version" is based on the actual date.   
3. Include now if not too difficult: Currently, CMOR obtains the `Conventions` attribute value from one of the CMIP7 cmor variable tables (e.g., CMIP_atmos.json). I think it is set to CF-1.12 there, but I think some modeling groups might need to make use of CF features in CF-1.13.  The CMIP7 specs say that `Conventions` should be one of the following:  "CF1.11", "CF1.12", or "CF1.13".  Therefore to accommodate these options, I think we should implement the recommendation regarding `Conventions` I made in https://github.com/PCMDI/cmor/issues/936
4. Include now or in a future release: I think we should remove the "tracking_prefix" and `drs_specs` from the user's input json file and instead rely solely on the value specified in the [cmor_cvs.json](https://github.com/WCRP-CMIP/cmip7-cmor-tables/blob/main/tables-cvs/cmor-cvs.json) file.  See some relevant discussion at https://github.com/PCMDI/cmor/issues/936 .

mauzey1:
Resolves #937 

Adds CMIP7-specific requirements for parent experiments (`parent_experiment_id`) and other parent attributes.  These are enabled by setting `"_cmip7_option": 1` in the user input JSON.

For experiments with no parent experiments defined in the CV, CMOR now warns on stray parent attributes and removes them from the dataset before writing. It also hard-errors when a user supplies `parent_experiment_id` for an experiment that should not have one, and when a required `parent_experiment_id` is missing for an experiment that does define parents.

Test/test_cmor_parent_attrs.py covers the following cases:

  - CMIP7 output with no parent attributes for no-parent experiments.
  - CMIP7 output with the full required parent attribute set for experiments like historical.
  - Warning-and-removal behavior when stray parent attributes are present without a parent experiment.
  - Erroring when a user defines `parent_experiment_id` for an experiment whose CV lists no parents.
  - Erroring when a required `parent_experiment_id` is omitted.
  - Erroring when the user supplies a parent experiment not allowed by the CV.
  - Erroring when the user supplies an incorrect `parent_activity_id`.

@matthew-mizielinski


taylor13:
The changes listed above seem sufficient to handle everything the way we want.  I've looked at the code changes briefly and didn't notice any issues, but can't say that should be a consideration.

durack1 (approved):
@mauzey1 looks good to me.

@matthew-mizielinski did you have the appetite to take a peek at this - honestly, it might be quicker to get the release done, and check this once live

PR #950, opened 2026-04-21
Fix file names for CMIP7 climatology datasets

### LINKED ISSUE #939: Error in creating dateRangeDD in filename
@matthew-mizielinski: do you see anything wrong with what I'm suggesting here?

In [this section of code](https://github.com/PCMDI/cmor/blob/main/Src/cmor_CV.c#L1617-L1622), a change is needed because in CMIP7 we handle frequency differently from CMIP6.  The "frequency_code" assigned in this section of code is used to determine the precision of the "time range" dates that appear at the end of each filename.  To handle this correctly in CMIP7, we'll need to replace

```
        if (strstr(frequency, "yr") != NULL) {
            frequency_code = 1;
        } else if (strstr(frequency, "dec") != NULL) {
            frequency_code = 1;
        } else if (strstr(frequency, "monC") != NULL) {
            frequency_code = 6;

```
With something like:

```
        if (cmor_tables[0].axes[timeDim].climatology == 1) != NULL) {
            frequency_code = 6;
        } else if (strstr(frequency, "yr") != NULL) {
            frequency_code = 1;
        } else if (strstr(frequency, "dec") != NULL) {
            frequency_code = 1;

```

It's important to identify "climatology" in a different way and to handle before monthly and higher frequency cases.  If you want to keep the code backward compatible the first if statement in the revised version above will have to include an "or" construct which in essence looks like:

       if (climatology=1 .or. frequency="monC"), frequency_code=6.  

but you'll have to use the correct syntax and names.  For CMIP7 there is no monC frequency, so that won't impact CMOR's processing of CMIP7.

mauzey1:
Fixes #939 

This PR enables the use of the `monC` time range for the file names of CMIP7 climatology datasets.

The removal of the appended `-clim` for CMIP7 datasets was performed in #949.

### DIFF
--- a/Src/cmor.c
+++ b/Src/cmor.c
@@ -6599,12 +6599,15 @@ int cmor_build_outname(int var_id, char *outname ) {
     int i,j;
     int n;
 
+    const int time_axis_id = cmor_vars[var_id].axes_ids[0];
+    const cmor_axis_t *time_axis = &cmor_axes[time_axis_id];
+    const cmor_axis_def_t *time_axis_def = 
+            &cmor_tables[time_axis->ref_table_id].axes[time_axis->ref_axis_id];
+
     /* -------------------------------------------------------------------- */
     /*      ok at that point we need to construct the final name!           */
     /* -------------------------------------------------------------------- */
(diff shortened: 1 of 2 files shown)


taylor13 (approved):
Looks good.  Thanks!

PR #960, opened 2026-05-01
Allow user override of `Conventions` attribute

### LINKED ISSUE #935: "product" and "Conventions"
@mauzey1 @matthew-mizielinski @JamesAnstey, @durack1: Both the `product` and `Conventions` global attributes are currently obtained from the header of the CMOR variable tables.  We would like variable tables to serve (without modification) other projects (e.g., obs4MIPs), which might produce a different `product` or might recommend a different version of the CF conventions.  Furthermore, for CMIP7 I think we've decided that data providers could choose to set `Conventions` to either CF-1.11, CF-1.12, or CF-1.13.  

I think currently CMOR might already allow the the table values for these to be overridden by users.  Is that true?  

In my view, Ideally, 
1. Any needed code revisions won't be difficult to implement.
2. CMOR would remain able to read older files (I don't think this is essential though)
3. We could specify the `product` and `Conventions` attribute in the project-specific CMOR CVs file, which would override what was in the variable table header.
4. Users could specify `Conventions` (but *not* `product`) in the CMOR (user-) input file, which would override any other settings.

For CMIP7 we would then set `product` to "model output" in the CMOR CVs file and set the default value of `Conventions` to "CF-1.12" in the CMOR CVs file.  The `Conventions` could be overridden by users in their CMOR input file.  The `product` and `Conventions` could be removed from the CMOR variable tables.

What do you think?

mauzey1:
Resolves #935 

This will allow users to set the `Conventions` attribute from their input JSON file. If the user doesn't provide a `Convention` attribute, then CMOR will get the one provided by the variable table.

This is similar behavior that is currently supported by for the `product` attribute.


TestTables/CMIP7_CV.json
```diff
@@ -1,5 +1,10 @@
 {
     "CV":{
+        "Conventions": [
+            "CF-1.11",
+            "CF-1.12",
+            "CF-1.13"
```
durack1, line 6: QC checking is synced with this list to-date, see https://esgvoc.ipsl.fr/concept/conventions?projects=cmip7. Would be good to connect if QC changes occur over the phase

durack1 (approved):
Looks good and syncs with https://esgvoc.ipsl.fr/concept/conventions?projects=cmip7, which is captured in [WCRP-CMIP/cmip7-cmor-tables/tables-cvs/cmor-cvs.json](https://github.com/WCRP-CMIP/cmip7-cmor-tables/blob/fb2b4324a4673ebb6aab52cc49cca79fe6a5b5e1/tables-cvs/cmor-cvs.json#L3-L7)

PR #970, opened 2026-05-13
Add the table's file name to the `table_info` string.

### LINKED ISSUE #964: Do we want to include "table_info" in files?
table_info isn't a CMIP7-recognized global attribute and it doesn't include the name of the table file.  Do we really want to include the following?

		:table_info = "Creation Date:(2026-04-21 15:01:29) MD5:bafb8e42cac0efce87531baf8e7d21b2" ;

If we do, I suggest adding at the beginning the name of the table.

mauzey1:
Resolves #964

### DIFF
--- a/Src/cmor.c
+++ b/Src/cmor.c
@@ -3285,7 +3285,8 @@ int cmor_setGblAttr(int var_id)
 /* -------------------------------------------------------------------- */
 /*    Set attribute Table_Info for netCDF file (CMIP6)                  */
 /* -------------------------------------------------------------------- */
-    snprintf(msg, CMOR_MAX_STRING, "Creation Date:(%s) MD5:",
+    snprintf(msg, CMOR_MAX_STRING, "Name: %s; Creation Date:(%s) MD5:",
+             cmor_tables[nVarRefTblID].path,
              cmor_tables[nVarRefTblID].date);
 
     for (i = 0; i < 16; i++) {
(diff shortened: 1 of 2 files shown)


taylor13 (changes requested):
Does this write the full path or just the table name.  I think to keep the global attribute listing less cluttered the filename would be sufficient, but if that's difficult, the full path would be acceptable.  Just say so, and I'll "approve" instead of "Request changes".  Thanks.

mauzey1:
@taylor13 The "path" of a table is usually the file name of the table, which matches what you had in https://github.com/PCMDI/cmor/issues/964#issuecomment-4433248715. I'm not sure if it ever includes other parts of the file path.  Alternatively, I can use "szTable_id" so that it uses the table's `table_id` instead.

taylor13:
Not concerned about how it's coded.  If the table name will appear preferably without the directory path in front of it, that's ideal.  Thanks.

taylor13:
I'm not sure how I can now officially approve, if that's needed.  Let me know if I should do anything.
thanks,
Karl

mauzey1:
@taylor13 I've updated the code so that it will just get the base name without any trailing path.

You can approve the pull request by going to "Files changed" tab, clicking on the "Submit review" button, and checking "Approve" and submitting the review.

taylor13 (approved):
Thanks!

PR #972, opened 2026-05-18
Ignore interval checks for climatology time axes, and fix interval checks being ingored for time1 axes

### LINKED ISSUE #943: approx_interval
This may or may not be an issue.  Nearly all time axes have approximately evenly space time-coordinates (monthly, daily, 6-hourly, etc.).  I think CMOR checks that within a file the time samples reported have coordinates consistent with that by checking the interval between successive time values against `approx_interval` (in cmor_axes.c?).  

In CMIP7 there is one time axis where the time coordinates are not approximately evenly spaced: time3.  This is used in reporting the monthly mean diurnal cycle (i.e. 24 values spaced hourly, but only 1 set of values per month).  The time coordinates (in units of days) thus look something like  15.0208, 15.0625, 15.1042, ... 15.9792, 45.0208, 45.0625, .... 45.9792 .....  So, the samples are spaced hourly for the first month then skip to the middle of the next month and again spaced hourly.  

For this case, I think we need to skip checking whether the values have been correctly reported.    If I'm right you can identify the axis to skip if it is "time3" _or_ if it is a "climatology" *and* "1hr" frequency.

mauzey1:
Resolves #943 

These changes will allow CMOR to write datasets for climatology variables, specifically for variables that have the "Diurnal Mean" time axis `time3`. It will allow them by ignoring the time interval check since the diurnal mean time axes will have irregular intervals.

This pull request will also fix a bug where the interval check for `time1` axes are ignored when they are not defined with bounds values using `cmor_axis`.

### DIFF
--- a/Src/cmor_axes.c
+++ b/Src/cmor_axes.c
@@ -2210,31 +2210,34 @@ int cmor_axis(int *axis_id, char *name, char *units, int length,
                                         length,
                                         name
                                         );
+            }
+        }
+
+        if ((refaxis.axis == 'T') && (refaxis.climatology == 0)) {
 /* -------------------------------------------------------------------- */
 /*      here we check if interval is about right                        */
 /*      just keep the begining of units out                             */
 /*      no need to know the since par                                   */
 /* -------------------------------------------------------------------- */
-                j = 0;
-                while (refaxis.units[j] == ' ')
-                    j++;
-                i = 0;
-                while ((refaxis.units[i + j] != ' ') &&
-                       (refaxis.units[i + j] != '\0')) {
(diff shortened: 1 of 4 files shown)


taylor13 (approved):
Looks good (but I didn't check the "test" code). Will we now bypass the check on interval for *all* climatology axes, or only the time3 axis?  Either way is o.k., but I'd like to know.

thanks,
Karl

mauzey1:
@taylor13 Using `cmor_write` to pass time values for climatology variables, not just those with `time3`, will now skip the time interval check just like when passing time values to `cmor_axis`.

taylor13:
Great!  Thanks.

Issue #837, opened 2025-07-07
preparing a CMOR demo with a coordinate having character string entries


gleckler1:
@mauzey1 @durack1 @taylor13 
With the goal of setting up a cmor demo for obs4MIPs, I am trying to reproduce a [CMIP6 variable](https://github.com/PCMDI/cmip6-cmor-tables/blob/main/Tables/CMIP6_Omon.json#L1835-L1852) that has dimensions [latitude basin time], with "basin" being an array of character strings.   A copy of the input file (hfbasin_Omon_UKESM1-0-LL_ssp370_r5i1p1f2_gnz_201501-204912.nc) used in this demo is in the same GH directory as the [test.py script](https://github.com/PCMDI/obs4MIPs-cmor-tables/blob/480_pjg_insitcollection/demo/demo-site_colletion/test.py).  [It appears to have been prepared with CMOR](https://github.com/PCMDI/obs4MIPs-cmor-tables/blob/480_pjg_insitcollection/demo/demo-site_colletion/ncdump.txt) as a contribution to CMIP6.  

The difficulty I am having is almost certainly with data type of sector
cmorBasin = cmor.axis("basin", coord_vals=sector[:],units="")

but there is no error until the cmor.write (below) and no other info is provided.

Any hints???

(xarray-cmor-v20250616) u7510@perlmutter:login14:~/git/obs4MIPs-cmor-tables/demo/demo-site_colletion> python -i test.py 
Traceback (most recent call last):
  File "/global/u1/u/u7510/git/obs4MIPs-cmor-tables/demo/demo-site_colletion/test.py", line 42, in <module>
    cmor.write(varid,d,len(time))
    ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "/global/homes/u/u7510/.conda/envs/xarray-cmor-v20250616/lib/python3.13/site-packages/cmor/pywrapper.py", line 799, in write
    raise Exception("Error could not convert data to a numpy array")
Exception: Error could not convert data to a numpy array

gleckler1:
Thanks, @mauzey1,  I have fixed the two issues you have identified and reproduced the segfault.  I suspect it is related to cmorBasin but unsure how that can be resolved.

mauzey1:
@gleckler1 The segfault is being caused by NaN values in the variable data.

Adding the line `d = np.where(np.isnan(d), 1.0e20, d)` to remove the NaNs will remove the segfault.

gleckler1:
[...]

Issue #832, opened 2025-06-19
Handling of grids other than lat-lon


taylor13:
A question was raised by @captainjasper [here](https://github.com/PCMDI/obs4MIPs-cmor-tables/issues/487) regarding demo codes for polar stereographic projection grids.  I'll copy it here to reach an additional audience.  Here's what it says:

> We are trying to prepare a new data set of surface elevation change of the Greenland Ice Sheet for obs4MIPs using the CMOR package in Python. Given the location of our data, we are using a polar stereographic projection (EPSG: 3413) instead of the recommended lat-lon grid since it would cause a number of issues close to the North Pole.
> 
> Are there any groups that have already used CMOR to handle grids other than lat-lon and would be willing to share some insights on how to best adjust the demo script for this case? Or other groups planning to register/process data sets in non-lat-lon grids who would like to team up on the issue?

I think folks in CORDEX use CMOR to write data on different projections, so maybe someone there can help.

durack1:
@captainjasper any chance we could get pointed at some example data to build this example?

ping @gleckler1, @TomLav - this is relevant to obs4MIPs discussions from deep time (e.g., https://github.com/PCMDI/obs4MIPs-cmor-tables/issues/471 and similar)

TomLav:
Hello Paul,

Certainly ! Example sea-ice concentration fields for both Northern Hemisphere and Southern Hemisphere polar grids (EASE2) are online e.g. here:

- NH : https://thredds.met.no/thredds/catalog/osisaf/met.no/reprocessed/ice/conc_450a1_files/monthly/2020/catalog.html?dataset=osisaf/met.no/reprocessed/ice/conc_450a1_files/monthly/2020/ice_conc_nh_ease2-250_cdr-v3p1_202002.nc
- SH : https://thredds.met.no/thredds/catalog/osisaf/met.no/reprocessed/ice/conc_450a1_files/monthly/2020/catalog.html?dataset=osisaf/met.no/reprocessed/ice/conc_450a1_files/monthly/2020/ice_conc_sh_ease2-250_cdr-v3p1_202002.nc

Both are for a monthly mean SIC in Feb 2020. We have all the other months in similar format online when needed.

Thanks for your effort and let me know if I can help.
Thomas

mauzey1:
We now have a CMOR example of handling the non lat-lon coordinates in the EASE2 data:
[...]

PR #976, opened 2026-06-02
Example CMIP7 cmorizing notebooks

### LINKED ISSUE #878: CMORizing notebook
I've transferred some emails here and hope further discussion will follow here.

***************
2/2/24
Hi PD, cc: P
 
I mentioned CMORising notebooks earlier. Piotr has thrown them into gists;
https://gist.github.com/piotr-florek-mohc/071256205f9e5f055f8b523a52694a01
https://gist.github.com/piotr-florek-mohc/02b40a8b9683814027d349bff4dd5bd5
https://gist.github.com/piotr-florek-mohc/92dbc05da9722497562ec740346b76fc
https://gist.github.com/piotr-florek-mohc/896c1a67cdb437d0f9447bfd8f582a74

(one of the above is a code library).
 
Theses were specifically for tidying up data for the WMO decadal forecasting centre and would need adapting for more general use, but it is a good way of providing an example.
 
Piotr hasn’t touched ocean variables in this form, so we still need to dig out the function call list for the 2D lat-lon variable production you asked about earlier in the week.
 
Adding something like this to the CMOR documentation or repo might be a nice way of making it a lot easier for people. If an appropriate set of notebooks could be constructed for Input4MIPs and Obs4MIPs then there might be less leaning on specific individuals for technical work/support.
 
M

***************
2/8/24
Thanks M, this is excellent – just looping in C, KT and PG.
 
The NEMO ocean I, j mapping would be really useful if that could be found, we are hoping to find a python template that we can use to write some non-lat/lon data, and the MOHC NEMO example looks like the kind of output we wanted to target.
 
Thanks again for this!
 
P

***************
2/8/24
FYI: P has been working on a 2D lat lon equivalent (i.e. CMIP6 Omon/tos) and will share when completed.
 
I would strongly recommend adapting these to act as demonstrations of how you can use CMOR in an adhoc fashion, and I think this route might well be simple enough that you could point obs4MIPs and input4MIPs contributors at them as a relatively simple DIY light weight CMORizer.
 
Matt

***************
12/19/24
Hi all,
 
We are just checking in to see if there might be any progress with this.  We don’t require anything polished – we can do that – just a clear-cut example of the few critical lines of CMOR code so that we can be confident it providing documentation for obs4MIPs data providers.   
 
Many thanks,
 PG

mauzey1:
Resolves #878 #832 #837

These notebooks show how to set up user info parameters, prepare data, and run CMOR to create NetCDF files for the CMIP7 project. They also show a visualization of the data.

These notebooks use truncated versions of real model data from CMIP5, CMIP6, etc.

There are examples that are relevant to #832 and #837.


durack1 (approved):
@mauzey1 this looks great!

You might want to add [nbstripout](https://anaconda.org/conda-forge/nbstripout) to the repo, so these notebooks are clean as a whistle as they are committed

PR #984, opened 2026-06-19
Use table file name instead of activity_id in variable comment

### LINKED ISSUE #979: Use of activity id in concatenated variable comments
This is not a big deal, but I noticed that if I pass a `comment` argument to `cmor.variable()` when there is already a `comment` for the variable in its CMOR table entry, in the netcdf file I get something like:
```
tas:comment = "this is my own comment, CFMIP_table_comment: this is the CMOR table comment"
```
where the `CFMIP` appears because this is my `activity_id` in this case. I *think* this is due to [this code](https://github.com/PCMDI/cmor/blob/2b1e7fc6ba34f9c870927c1bef163f5235bed9c4/Src/cmor_variables.c#L1270). Keeping both comments (from the user and from the CMOR table) is useful behaviour - my only question is why is the CMOR table comment labelled with the `activity_id` instead of, say, writing "CMOR_table_comment"? Given that in CMIP the `activity_id` is associated with the experiment, not the variable's provenance. @mauzey1 @taylor13

mauzey1:
Resolves #979

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -245,6 +245,7 @@ test_python: python
 	env TEST_NAME=Test/test_cmor_associated_variable.py make test_a_python
 	env TEST_NAME=Test/test_cmor_frequency_required.py make test_a_python
 	env TEST_NAME=Test/test_cmor_parent_attrs.py make test_a_python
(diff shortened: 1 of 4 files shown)


durack1 (approved):
Wonderful, thanks @mauzey1

Issue #983, opened 2026-06-12
Error message coloring format changes


mauzey1:
CMOR's error/warning messages get printed to the terminal with color formatting that makes letters red for errors or blue for warnings.  These messages then get printed to a white background.  I ran into an issue where sometimes the letters will be colored white causing them to become invisible.  This was being caused by how the color formatting characters interacted with characters in the error/warning message.

I am currently reworking how the error messages are formatted for printing to the terminal.

Another minor issue being worked on is the inclusion of the color formatting characters in the log file for CMOR.  Ideally, this formatting should not be included in log files.

ilaflott:
please make the color configurable for users! i dislike the color.

`CMOR_COLOR_OUTPUT`?

durack1:
> please make the color configurable for users! i dislike the color.

Hi @ilaflott we're in a bugfix mode to finalise CMOR3 for CMIP7, so not eager to make such a change at this late stage.

To make colour changes with the current code base requires a C recompile. A user-configurable setting would require a config file, an environment variable scan, or changes to `cmor.setup()` to be implemented, which is currently out of scope for CMOR3 finalisation, but definitely a good idea for CMOR4 future development

mauzey1:
Resolved by #985

PR #985, opened 2026-06-22
Check the CV file for strings that exceed CMOR's character limit

### LINKED ISSUE #982: No error or warning for CV string values exceeding 1023 character limit
Related to https://github.com/WCRP-CMIP/cmip7-cmor-tables/issues/112.

CMOR crashes if a CV string value exceeds the 1023 character limit.  No error or warning message is produced that could indicate which value is over the limit.

A warning or error should be thrown when strings go over the limit.  An error can simply show the user which attribute in the file is over the limit.  A warning could also be used to indicate that a string is over the limit and will be truncated to fit within the character limit.

I'm more in favor with the error approach since CMOR should keep dataset attributes true to the CV.

mauzey1:
Resolves #982, #983 

This adds the ability to check for strings in the CV JSON file that exceed CMOR's 1023 character limit for string values. This is to prevent CMOR from copying values that are too large for its fixed 1024 bytes long strings, which caused crashes.

This also fixes issues with the error messages' color formatting. Error messages printed to log files will no longer have color formatting characters.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -246,6 +246,7 @@ test_python: python
 	env TEST_NAME=Test/test_cmor_frequency_required.py make test_a_python
 	env TEST_NAME=Test/test_cmor_parent_attrs.py make test_a_python
 	env TEST_NAME=Test/test_cmor_variable_attr_comment.py make test_a_python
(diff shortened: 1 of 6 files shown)


durack1 (approved):
@mauzey1 wonderful, thanks!

PR #993, opened 2026-08-07
Update CMIP7 tables and examples

mauzey1:
This pull request updates the CMIP7 tables and removes the dummy source_id from the CMIP7 tests and examples.

It also updates the Python and Fortran examples, and adds C examples.


durack1 (approved):
All looks great to me

PR #995, opened 2026-08-25
Allow user-defined values for `source`, `experiment`, and `institution`

### LINKED ISSUE #988: Handling of optional global attributes
I'm moving here the part of the discussion In https://github.com/WCRP-CMIP/cmip7-cmor-tables/issues/129 regarding handling of the `source` attribute, which is optional in CMIP7 (with its value uncontrolled).  

As I understand it, CMOR gets its value for `source` from the cmor-cvs.json file, and the user can't override the value in the cmor-cvs.json file (if I correctly understand https://github.com/WCRP-CMIP/cmip7-cmor-tables/issues/129#issuecomment-4836008081). I suggest that CMOR should treat `source` as follows:

- by default, extract for `source` whatever is recorded as `source` in cmor-cvs.json and write it as the `source` global attribute. (I think that is what happens now.). 
- Enable users to override the `source` value found in cmor-cvs.json by specifying a different value in the *_input.json file.  More specifically, if in the user's input file (e.g., https://github.com/PCMDI/cmor/blob/main/Test/CMOR_input_example.json) the user specified a value (other than "") for `source`, then CMOR would override the value provided by the cmor-cvs.json file.

We might want to do something similar for `institution` and `experiment`.

mauzey1:
Resolves #988 

Allow the user to set values for `source`, `experiment`, and `institution`.

This will overwrite the values that are set by default from the CV for `source_id`, `experiment_id`, and `institution_id`.

If `source`, `experiment`, and `institution` are in `required_global_attributes`, then the values from the CV will overwrite the user values.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -237,6 +237,7 @@ test_python: python
 	env TEST_NAME=Test/test_cmor_unsupported_calendar.py make test_a_python
 	env TEST_NAME=Test/test_cmor_time_interval_check.py make test_a_python
 	env TEST_NAME=Test/test_cmor_license_attributes.py make test_a_python
+	env TEST_NAME=Test/test_cmor_optional_derived_attributes.py make test_a_python
 	env TEST_NAME=Test/test_cmor_nested_cv_attribute.py make test_a_python
 	env TEST_NAME=Test/test_cmor_path_and_file_templates.py make test_a_python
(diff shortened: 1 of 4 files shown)


durack1 (approved):
@mauzey1 looks good to me!

